PR #3, opened 2015-06-06
Add isGone method to \Zend\Http\Response

asgrim:
As per the discussion in zendframework/zf2#7368 this change adds a new
method \Zend\Http\Response->isGone() to determine if the status code is
a "410 Gone". This works exactly the same as the similar methods like
isNotFound simply by checking the status code.

This change takes the opinion that a 410 Gone should not be included in
the isNotFound() method. Further discussion can be seen in the issue.

This was originally opened on zendframework/zf2#7536.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -400,6 +400,16 @@ class Response extends AbstractMessage implements ResponseInterface
     }
 
     /**
+     * Does the status code indicate the resource is gone?
+     *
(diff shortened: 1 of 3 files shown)


mwillbanks:
@asgrim could you please change the target of this to the develop branch?

asgrim:
Gone to #11

Issue #6, opened 2015-06-26
Call to undefined method Zend\Http\Response::extractCode


neilime:
Hi,

Zend\Http\Client\Adapter\Proxy is calling an undefined method, how could it works ?

https://github.com/zendframework/zend-http/blob/master/src/Client/Adapter/Proxy.php#L247

Maks3w:
Fixed https://github.com/zendframework/zend-http/commit/886274e4176c42daa6515712e39fd3d62bc4afe6

PR #7, opened 2015-06-26
Fix "Call to undefined method Zend\Http\Response::extractCode"

neilime:
https://github.com/zendframework/zend-http/issues/6


weierophinney:
Indeed, fixes one of the online tests! Thanks!

PR #10, opened 2015-07-08
FIX - Put back the insensitive replacements for Transfer-Encoding and Content-Encoding.

nbochenko:


nbochenko:
@blanchonvincent

Martin-P:
@nbochenko Perhaps you can add some text to this PR and:
- explain what bug you encountered
- create unit tests to reproduce the issue
- explain why your code change is a fix for your issue

weierophinney:
@nbochenko Ping - Please see the comments from @Martin-P ; we need more detail, specifically in the form of unit tests, to ensure that the changes are correct, and that we don't accidentally revert them in the future.

Maks3w:
Closed due inactivity

wiese:
I get your point wanting a test for this, so we don't break it again in the future, but this fixes a real bug introduced during https://github.com/nbochenko/zend-http/commit/ddf5a83c60120450cc15467a886192de4f2c229f, which noboby had any issue merging into the main line without a test - so some initiative from the core team and the original author should be just as expected as from someone who, at least, found the bug and was kind enough to report it. I'll try and contribute some coverage here rather sooner than later.

Martin-P:
Ping @Maks3w  @weierophinney.
I can confirm what @wiese says. @kasitmp never added any tests to the original PR (https://github.com/zendframework/zf2/pull/6691) and @weierophinney just merged the PR without tests.

Before https://github.com/zendframework/zf2/pull/6691 there was `stri_replace` (watch the **i**), but that PR removed the case-insensitive matching and made it case-sensitive.

Perhaps @kasitmp or @weierophinney can add the proper tests for this as they caused this issue :wink:

PR #8, opened 2015-08-05
Missing setting of CURLINFO_HEADER_OUT option

weierophinney:
The CURLINFO_HEADER_OUT option is not set in the curl adapter which results in Client::getLastRawRequest to be empty

### DIFF
--- a/src/Client/Adapter/Curl.php
+++ b/src/Client/Adapter/Curl.php
@@ -348,6 +348,9 @@ class Curl implements HttpAdapter, StreamInterface
         curl_setopt($this->curl, CURLOPT_HTTP_VERSION, $curlHttp);
         curl_setopt($this->curl, $curlMethod, $curlValue);
 
+        // Set the CURLINFO_HEADER_OUT flag so that we can retrieve the full request string later
+        curl_setopt($this->curl, CURLINFO_HEADER_OUT, true);
+
         if ($this->outputStream) {
             // headers will be read into the response
             curl_setopt($this->curl, CURLOPT_HEADER, false);


weierophinney:
I've converted this issue into a pull request, and it provides a fix. I have *zero* idea how to test this other than doing an integration test; @bashofmann or @benjaminradtke - can one of you test with one of your apps?

weierophinney:
I enabled the online tests locally, and discovered that adding this *fixes* several existing test failures.

I'll get this merged and released ASAP. Thanks for the report and the clarifications!

benjaminradtke:
Hey @weierophinney thanks for your quick reply. Of course I will test it with our app, to remove our workaround.

benjaminradtke:
Works like a charm - but I am not sure if you should put that statement in the later else path:

    if ($this->outputStream) {
        // headers will be read into the response
        curl_setopt($this->curl, CURLOPT_HEADER, false);
        curl_setopt($this->curl, CURLOPT_HEADERFUNCTION, [$this, "readHeader"]);
        // and data will be written into the file
        curl_setopt($this->curl, CURLOPT_FILE, $this->outputStream);
    } else {
        // ensure headers are also returned
        curl_setopt($this->curl, CURLOPT_HEADER, true);
        curl_setopt($this->curl, CURLINFO_HEADER_OUT, true);

        // ensure actual response is returned
        curl_setopt($this->curl, CURLOPT_RETURNTRANSFER, true);
    }

This is, where it was placed before since ZF 1.12.8

weierophinney:
> but I am not sure if you should put that statement in the later else path

Considering we *always* call `curl_getinfo` later to retrieve the raw request, having it outside the conditional makes sense.

PR #14, opened 2015-08-05
PhpEnvironment\Request::detectBaseUrl() - E_WARNING: strpos(): Empty needle

weierophinney:
I think those checks on following lines https://github.com/zendframework/zend-http/blob/master/src/PhpEnvironment/Request.php#L498 should use ``empty()`` instead of ``!== null`` because in some cases it triggers warning on a line:
https://github.com/zendframework/zend-http/blob/master/src/PhpEnvironment/Request.php#L522

### DIFF
--- a/src/PhpEnvironment/Request.php
+++ b/src/PhpEnvironment/Request.php
@@ -515,6 +515,11 @@ class Request extends HttpRequest
             }
         }
 
+        // If the baseUrl is empty, then simply return it.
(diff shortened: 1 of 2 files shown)


weierophinney:
I've converted this issue with a pull request, which adds a test for the condition (it failed prior to the patch), and a patch to ensure the condition does not occur.

Issue #15, opened 2015-08-05
cookie values decode error


malinink:
Hi!
I have found one problem with encoding.

In PHP it's normal behavior that values are urlencoded, but in other languages there are another standards for encoding such value, for example - in python.

So, cookie created in python could be invalid after urldecode here:
https://github.com/zendframework/zend-http/blob/master/src/Header/SetCookie.php#L119

But if we want to fix that issue, we suppose to get rid of that urlencode also:
https://github.com/zendframework/zend-http/blob/master/src/Header/SetCookie.php#L236

That is wrong, cause when we try to set cookie with specific symbols - we get wrong cookie.

I really don't know how to fix problem, have anyone ideas?

I thought that best approach is to add client type (php, python) and add specific decode/encode functions for each type, also add default type which only encode values.

At now I have a site that causes Zend Http Client failed on send method:

``` php
        $client = new Client('http://www.gd.ru/articles/3514-uvelichenie-prodaj');
        $client->send();
```

malinink:
For example default action may looks like that one.

weierophinney:
@malinink Can you please submit your changes as a pull request?

malinink:
@weierophinney Pull request have been added.

zerocrates:
I mistakenly filed a duplicate issue for this.

I think the best way to distill it down is to post a tiny one-liner that fails when it shouldn't:

```php
Zend\Http\Header\SetCookie::fromString('Set-Cookie:foo=%0A; path=/; HttpOnly');
```

The practical upshot is that any attempt to make a *request* to a server that emits cookies with urlencoded whitespace in the value will fail because the SetCookie header class is urldecoding the value before attempting to validate it.

PR #16, opened 2015-08-05
fix bug with urldecoding cookie value

malinink:
Discussion could be found in issue #15

### DIFF
--- a/src/Header/SetCookie.php
+++ b/src/Header/SetCookie.php
@@ -116,7 +116,7 @@ class SetCookie implements MultipleHeaderInterface
                     // First K=V pair is always the cookie name and value
                     if ($header->getName() === null) {
                         $header->setName($headerKey);
-                        $header->setValue(urldecode($headerValue));
+                        $header->setValue($headerValue);
                         continue;
                     }
 
--- a/test/Header/SetCookieTest.php
+++ b/test/Header/SetCookieTest.php
@@ -557,7 +557,7 @@ class SetCookieTest extends \PHPUnit_Framework_TestCase
                     'secure'  => true,
                     'httponly'=> false
                 ],
-                'PHPSESSID=123456789+abcd%2Cef; Expires=Tue, 21-Nov-2006 08:33:44 GMT; Domain=.localdomain; Path=/foo/baz; Secure'
+                'PHPSESSID=123456789%2Babcd%252Cef; Expires=Tue, 21-Nov-2006 08:33:44 GMT; Domain=.localdomain; Path=/foo/baz; Secure'
             ],
             [
                 'Set-Cookie: myname=myvalue; Domain=docs.foo.com; Path=/accounts; Expires=Wed, 13-Jan-2021 22:23:01 GMT; Secure; HttpOnly',


weierophinney:
I'm not sure this makes sense.

The [specification indicates that the cookie values may consist of a cookie-octet](https://tools.ietf.org/html/rfc6265#section-4.1.1), as defined below:

```
cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
```

This clearly allows `+` *without encoding* (it's character 2B in the US ASCII set). If we encode by default, we then have to assume that clients receiving the cookie are capable of, and *will*, detecting and decoding the value, which is not a safe assumption.

Since we do not currently do any encoding, I would argue that if the client you're sending the cookie to requires encoding, you should do it manually when you set the cookie value. This approach keeps the current class flexible enough to handle both urlencoded and non-urlencoded values, and allowing it to encompass the entire spec, and not a subset of it.

malinink:
The main problem is that I don't want to send cookie, I just want to get them from remote.
And that operation failed cause cookie format is not php urlencode() function result.

For example, value could be:
```
K%B42%B6%AA.%06%12J%25%259J%D6%99V%86%26%26%86F%16%26f%C6%E6%D6%C5V%96VJ%A9e%A9y%25%C1%40%ACd%9Dde%00%1434%B2RJ%CE%CF%2B%01%0Ay%96%A4%E6%16%2BY%27Z%19%82%CC%00I8%16%95d%26%E7%A4%C6%1B%9B%1A%9A%80L3%B6%AE%AD%05%00
```

This is valid value, but after decoding it starting contain wrong symbols

And I catch Exception here:
https://github.com/zendframework/zend-http/blob/master/src/Header/SetCookie.php#L305

malinink:
@weierophinney 
Okay, I catch everything, now I can tell the main reason why current code is not correct.

Here https://github.com/zendframework/zend-http/blob/master/src/Header/SetCookie.php#L119
We set value of cookie from headers, but not real value, we urldecode it. In setter we check where cockie value is valid.

But, we must check value from headers _before_ we urldecode them! Cause in specification there is no one world about encode/decode functions. This is only PHP specific way to store them.

zerocrates:
It seems to me that @malinink is right here about there being a problem: the cookie value gets urldecoded when reading in from a string on line 119, and then re-encoded on the way out on line 236.

The problem with this PR is that by now, that automatic decoding and encoding is expected from this class: it's been the behavior for years. So, you can't just take out the urldecode on line 119 and be okay.

The least-invasive solution I can come up with is to `urlencode` the value passed to `HeaderValue::assertValid` on [line 305](https://github.com/zendframework/zend-http/blame/master/src/Header/SetCookie.php#L305). This won't change the input or output from this class at all, it will just make it so that the validation check is working against the correct data, the same data that will be output by `getFieldValue` on line 236.

The downside is that you do a little useless round-trip between urldecode and urlencode, but the upside is that it should merely fix the problem without altering the external behavior.

PR #23, opened 2015-09-11
Fix invalid exception from urlencoded cookie value

### LINKED ISSUE #15: cookie values decode error

zerocrates:
This is my alternative solution to #15, as a replacement for the existing PR #16.

This PR fixes the problem by simply re-urlencoding the cookie value before validating it. This is allowable because the actual cookie as output *also* urlencodes the value (see [SetCookie::getFieldValue](https://github.com/zerocrates/zend-http/blob/master/src/Header/SetCookie.php#L236).

As a concrete example, this PR removes the test `testPreventsCRLFAttackViaConstructor`. The change makes that test fail, but the test was already failing for an input that wouldn't have actually enabled a CRLF attack. The "evil" example is:

```php
$header = new SetCookie("leo_auth_token", "example\r\n\r\nevilContent");
```

However, under both the existing code and after the commit in this PR, that SetCookie header creates no CRLF injection problems, because SetCookie already automatically urlencodes values on output. So, the output of `$header->toString()` on the above "evil" cookie is already safe:

```
Set-Cookie: leo_auth_token=example%0D%0A%0D%0AevilContent
```

I think this change is the best way to fix the regression without compromising the security the validation was introduced to provide and without surprising users by changing the header's behavior.

Fix #15
Supersede & close #16

### DIFF
--- a/src/Header/SetCookie.php
+++ b/src/Header/SetCookie.php
@@ -302,7 +302,6 @@ class SetCookie implements MultipleHeaderInterface
      */
     public function setValue($value)
     {
-        HeaderValue::assertValid($value);
         $this->value = $value;
         return $this;
     }
--- a/test/Header/SetCookieTest.php
+++ b/test/Header/SetCookieTest.php
@@ -9,6 +9,7 @@
 
 namespace ZendTest\Http\Header;
 
+use Zend\Http\Header\HeaderValue;
 use Zend\Http\Header\SetCookie;
 
 class SetCookieTest extends \PHPUnit_Framework_TestCase
@@ -431,8 +432,8 @@ class SetCookieTest extends \PHPUnit_Framework_TestCase
      */
     public function testPreventsCRLFAttackViaConstructor()
     {
-        $this->setExpectedException('Zend\Http\Header\Exception\InvalidArgumentException');
         $header = new SetCookie("leo_auth_token", "example\r\n\r\nevilContent");
+        $this->assertTrue(HeaderValue::isValid($header->getFieldValue()));
     }
 
     public function setterInjections()
--- a/test/Header/SetCookieTest.php
+++ b/test/Header/SetCookieTest.php
@@ -435,6 +435,13 @@ class SetCookieTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('Set-Cookie: leo_auth_token=example%0D%0A%0D%0AevilContent', $header->toString());
     }
 
+    public function testPreventsCRLFAttackViaSetValue()
+    {
+        $header = new SetCookie("leo_auth_token");
+        $header->setValue("example\r\n\r\nevilContent");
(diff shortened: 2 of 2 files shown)


src/Header/SetCookie.php
```diff
@@ -302,7 +302,7 @@ public function getName()
      */
     public function setValue($value)
     {
-        HeaderValue::assertValid($value);
+        HeaderValue::assertValid(urlencode($value));
```
Maks3w: When this won't be valid? if there no case exists then remove whole line.

test/Header/SetCookieTest.php
```diff
@@ -425,16 +425,6 @@ public function testPreventsCRLFAttackViaFromString()
         $header = SetCookie::fromString("Set-Cookie: leo_auth_token=example;\r\n\r\nevilContent");
     }
 
-    /**
```
Maks3w: Don't remove the test, Instead assert the value is valid. If not regressions are welcome :)

zerocrates:
I see, yeah that's a much better solution. I've replaced my initial commit with one that follows your suggestions.

test/Header/SetCookieTest.php
```diff
@@ -431,8 +432,8 @@ public function testPreventsCRLFAttackViaFromString()
      */
     public function testPreventsCRLFAttackViaConstructor()
     {
-        $this->setExpectedException('Zend\Http\Header\Exception\InvalidArgumentException');
         $header = new SetCookie("leo_auth_token", "example\r\n\r\nevilContent");
+        $this->assertTrue(HeaderValue::isValid($header->getFieldValue()));
```
Maks3w: I preffer if you assertEquals against the __toString() output. Will be more clear to understand because the encoded characters will be visible while reading this code.

Maks3w: Could you add an additional test method for unit check of `setValue`?

zerocrates:
Okay. (I assume the extra test method is what you were looking for, if not let me know)

Maks3w:
@weierophinney Please add this one to 2.4.8 LTS

weierophinney:
Tagged in release-2.5.2 and 2.4.8.

PR #25, opened 2015-10-01
Fix #21

### LINKED ISSUE #21: Zend/Http/Header/AbstractAccept throws php notice when "HTTP_ACCEPT_LANGUAGE" is non well formatted
Hi,

An error occured when we try to retrieve "Accept-Language" from request headers and "$_SERVER['HTTP_ACCEPT_LANGUAGE']" is non well formatted : ""Undefined offset: 1" - /.../Zend/Http/Header/AbstractAccept.php (173)"

The following unit test shows this behavior.

```php
    public function testDefineDefaultLocaleWithWrongHttpAcceptLanguageHeader()
    {
        $_SERVER['HTTP_ACCEPT_LANGUAGE'] = 'en-US;en';
        $oRequest = new \Zend\Http\PhpEnvironment\Request();
        $oRequest->getHeaders()->get('Accept-Language'); // Php notice
    }
```
Thanks for helping me to fix it

neilime:
Prevents non well formatted accept  header value.

Fix #21

### DIFF
--- a/src/Header/AbstractAccept.php
+++ b/src/Header/AbstractAccept.php
@@ -169,9 +169,9 @@ abstract class AbstractAccept implements HeaderInterface
 
             foreach ($paramsStrings as $param) {
                 if (strpos($param, '=') === false) {
-                    continue;    
+                    continue;
                 }
-                
+
                 $explode = explode('=', $param, 2);
 
                 $value = trim($explode[1]);
--- a/src/Header/AbstractAccept.php
+++ b/src/Header/AbstractAccept.php
@@ -168,6 +168,10 @@ abstract class AbstractAccept implements HeaderInterface
             }
 
             foreach ($paramsStrings as $param) {
+                if (strpos($param, '=') === false) {
+                    continue;    
+                }
+                
                 $explode = explode('=', $param, 2);
 
                 $value = trim($explode[1]);
--- a/test/Header/AcceptLanguageTest.php
+++ b/test/Header/AcceptLanguageTest.php
@@ -31,6 +31,12 @@ class AcceptLanguageTest extends \PHPUnit_Framework_TestCase
         $acceptLanguageHeader = AcceptLanguage::fromString('Accept-Language: xxx');
         $this->assertEquals('xxx', $acceptLanguageHeader->getFieldValue());
     }
+    
+    public function testAcceptLanguageGetFieldValueReturnsProperNonWellFormattedHeaderValue()
+    {
+        $acceptLanguageHeader = AcceptLanguage::fromString('Accept-Language: en-US;en');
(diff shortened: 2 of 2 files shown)


Freeaqingme:
Awesome!

As you can see above one of the CS checks failed, apparently on some trailing spaces. Something you could perhaps take a look at as well?

Results are here: https://travis-ci.org/zendframework/zend-http/jobs/83107792

neilime:
Done !

Freeaqingme:
@Maks3w Is this something you could look at? I've been out of the game for quite some time (although I still feel for this particular component), so am I little uncomfortable merging it. As far as I'm concerned it's ready to go!

Maks3w:
Well I have questions. 

1. Because an abstract class has been modified then any child test should have a test and not only Abstract-Language

2. After take a look to http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html I don't see those headers should share the same ABNF so I think the abstract class should be split with trait functions and be deprecated

Maks3w:
Errata: RFC2616 are expressed using [EBNF](https://tools.ietf.org/html/rfc822)

Accept-Header is defined individually on [RFC3282](https://tools.ietf.org/html/rfc3282)

Maks3w:
After reading RFC 7231 only the weight param (`;q=[0,0.001-1]`) is common to all those headers but the rest of parsing values are not.

This is the resume of ABNF rules from http://tools.ietf.org/html/rfc7231#appendix-D
```
Accept = [ ( "," / ( media-range [ accept-params ] ) ) *( OWS "," [OWS ( media-range [ accept-params ] ) ] ) ]
Accept-Charset = *( "," OWS ) ( ( charset / "*" ) [ weight ] ) *( OWS "," [ OWS ( ( charset / "*" ) [ weight ] ) ] )
Accept-Encoding = [ ( "," / ( codings [ weight ] ) ) *( OWS "," [ OWS ( codings [ weight ] ) ] ) ]
Accept-Language = *( "," OWS ) ( language-range [ weight ] ) *( OWS "," [ OWS ( language-range [ weight ] ) ] )

accept-params = weight *accept-ext
weight = OWS ";" OWS "q=" qvalue

OWS = Optional White Space
```

Freeaqingme:
@Maks3w How is this all relevant for this particular pull request? Fwiw, the implementations that are included in zend\http for the various accept header types come with own regexes to support and validate the different formats.

Of course we can always refactor this class and deprecate parts of it. But that seems way beyond the scope of this particular issue: Ensuring no notices are thrown on invalid input.

Maks3w:
Reading the RFC and the ABNF resume split languages with `;` is not allowed. Should be `,`. So the proposed solution is wrong. The header was malformed and there is nothing to fix.

Maks3w:
Closed due inactivity.

PR #34, opened 2015-12-08
Using http_build_query for application/x-www-form-urlencoded

ameoba32:
By standard "&" character is used for application/x-www-form-urlencoded submissions. If  separator is set to different value via ini_set('arg_separator.output', 'some other'); it will break http client with wrong request.

I suggest to hardcode "&" there


redheadedstep:
Update on when this will be merged into master @weierophinney ?  Anytime I run composer update, I have to manually update this file.

ezimuel:
@ameoba32 please merge https://github.com/ameoba32/zend-http/pull/3 that contains the unit test for your PR.

ezimuel:
I added the unit test during the merge.

Issue #50, opened 2016-02-22
Curl Adapter Needs to Look for Case Insensitive Headers


KarunaGovind:
\Zend\Http\Client\Adapter\Curl removes a few response headers (namely Transfer-Encoding, Content-Encoding) to make sure \Zend\Http\Response behaves correctly.

However, it assumes response headers have fixed cases. The AWS ELB in particular returns "transfer-encoding" as all lower case, therefore the code needs to do a case insensitive replace.

rsacksteder:
Ran into this exact same issue! Hopefully this can be reviewed and pulled into the next release.

samsonasik:
can be closed as it seems fixed at https://github.com/zendframework/zend-http/pull/53

PR #51, opened 2016-02-22
#50 - curl http header replacement needs to be case insensitive

KarunaGovind:
Ref ticket #50


ezimuel:
Closing this PR because overlap with #53.

Issue #57, opened 2016-05-16
Incorrect handling of accept headers


theboolean:
In the case the value of an accept header does not contain an equal sign, for example:
`Accept-Language: whatever; whateverElseWithoutEqualSign`
the following notice is triggered:
`Notice: Undefined offset: 1 in /path/zendframework/zend-http/src/Header/AbstractAccept.php on line 173`.

ezimuel:
Solved in #58

PR #58, opened 2016-05-16
Avoid triggering a notice with special crafted accept headers

theboolean:
Avoid triggering a notice when an accept header without an equal sign is parsed.

Refs #57

### DIFF
--- a/src/Header/AbstractAccept.php
+++ b/src/Header/AbstractAccept.php
@@ -170,7 +170,12 @@ abstract class AbstractAccept implements HeaderInterface
             foreach ($paramsStrings as $param) {
                 $explode = explode('=', $param, 2);
 
-                $value = trim($explode[1]);
+                if (count($explode) === 2) {
+                    $value = trim($explode[1]);
+                } else {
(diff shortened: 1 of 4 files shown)


Freeaqingme:
@theboolean Thanks for your Pull Request! Could you perhaps try to also include a test case as a unit test so we can be sure this bug won't resurface again, for example when someone refactors this class? Thanks!

theboolean:
Hi @Freeaqingme, I added the test, and as you can see I rewrote the history of my branch putting it before my patch, so it is clear that the test failed before it.
I also added another edge case test that passed also before applying my patch, just in case.

Freeaqingme:
Awesome!

@weierophinney Given how all tests pass this looks good to me, but I've been out of it for a while. So perhaps this is something you could look at as well?

ezimuel:
@theboolean Thanks for your contribution!

theboolean:
@ezimuel you're welcome!
If it looks good to you, you can also close the issue related to this PR (#57).

PR #59, opened 2016-06-02
move code-line

pine3ree:
$filename is only needed if the method does not short-circuit return due to an empty $baseUrl


ezimuel:
@pine3ree Thanks!

PR #60, opened 2016-06-10
fixed Content-Length header

### LINKED ISSUE #26: "Content-Length" header set to "0" returns header line without value
Hi

Setting the ```"Content-Length"``` header to ```0```, will return an empty header line ```Content-Length: ``` in the response. 
This might break CORS preflight requests processed by load-balancers (e.g. on fortrabbit).

Problem is located in the ```\Zend\Http\Header\ContentLength``` constructor, that does not sets the ```$value``` property if ```0``` is passed.

Issue can be fixed in ```\Zend\Http\Header\ContentLength``` either:

-  changing the constructor behavior (accepting ```0``` or ```'0'``` as value):
   from ```if ($value) {```  to ``` if ($value !== null) {```
   Note: all other classes in the Zend\Http\Header namespace have the same similar (buggy?) constructor (except GenericHeader).

- changing the ```getFieldValue``` method to:
  ``` {php}
    public function getFieldValue()
    {
        return $this->value ?: 0;
    }
```

As soon as fixed I recommend to change the requirement of "zfr/zfr-cors" to the newest version of "zendframework/zend-http"

Kind regards,
Curzio

### PR DESCRIPTION
**Pull request by DCMNMarc**
* It's no more possible to have a Content-Length header with `NULL` as value
* allow `Content-Length: 0`, fixes #26
* added `ContentLength::setFieldValue()`
* throw exception on invalid `Content-Type` headers like `Content-Type: 123.456`
* restructured unit tests

### DIFF
--- a/src/Header/ContentLength.php
+++ b/src/Header/ContentLength.php
@@ -38,12 +38,21 @@ class ContentLength implements HeaderInterface
         return $header;
     }
 
-    public function __construct($value = null)
+    public function __construct($value)
     {
-        if ($value) {
-            HeaderValue::assertValid($value);
-            $this->value = $value;
+        $this->setFieldValue($value);
+    }
+
+    public function setFieldValue($value)
+    {
+        $value = (string) $value;
+        if (!HeaderValue::isValid($value) || !preg_match('/^(0|[1-9][0-9]*)$/', $value)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                'Invalid header value for Content-Length: "%s"',
+                $value
+            ));
         }
+        $this->value = $value;
     }
 
     public function getFieldName()
--- a/test/Header/ContentLengthTest.php
+++ b/test/Header/ContentLengthTest.php
@@ -13,56 +13,89 @@ use Zend\Http\Header\ContentLength;
 
 class ContentLengthTest extends \PHPUnit_Framework_TestCase
 {
-    public function testContentLengthFromStringCreatesValidContentLengthHeader()
+    public function testFromStringCreatesValidContentLengthHeader()
     {
-        $contentLengthHeader = ContentLength::fromString('Content-Length: xxx');
+        $contentLengthHeader = ContentLength::fromString('Content-Length: 123');
         $this->assertInstanceOf('Zend\Http\Header\HeaderInterface', $contentLengthHeader);
         $this->assertInstanceOf('Zend\Http\Header\ContentLength', $contentLengthHeader);
     }
 
-    public function testContentLengthGetFieldNameReturnsHeaderName()
+    public function testGetFieldNameReturnsHeaderName()
     {
-        $contentLengthHeader = new ContentLength();
+        $contentLengthHeader = new ContentLength('123');
         $this->assertEquals('Content-Length', $contentLengthHeader->getFieldName());
     }
 
-    public function testContentLengthGetFieldValueReturnsProperValue()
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/Header/ContentLength.php [hunk 1]
```diff
@@ -38,12 +38,21 @@ public static function fromString($headerLine)
...
-    public function __construct($value = null)
+    public function __construct($value)
     {
-        if ($value) {
-            HeaderValue::assertValid($value);
-            $this->value = $value;
+        $this->setFieldValue($value);
+    }
+
+    public function setFieldValue($value)
```
ezimuel: Adding a `setFieldValue()` function is considered a new feature and it should be done with a new PR.

src/Header/ContentLength.php
```diff
@@ -38,12 +38,21 @@ public static function fromString($headerLine)
...
+    {
+        $value = (string) $value;
+        if (!HeaderValue::isValid($value) || !preg_match('/^(0|[1-9][0-9]*)$/', $value)) {
```
ezimuel: You can check for digit number using `false !== filter_var($value, FILTER_VALIDATE_INT)` instead of regex.

test/Header/ContentLengthTest.php
```diff
@@ -13,56 +13,89 @@
...
+        $contentLengthHeader = new ContentLength($power95of2);
+        $this->assertEquals($power95of2, $contentLengthHeader->getFieldValue());
+    }
 
     /**
-     * @see http://en.wikipedia.org/wiki/HTTP_response_splitting
-     * @group ZF2015-04
+     * @dataProvider invalidValues
      */
-    public function testPreventsCRLFAttackViaFromString()
```
ezimuel: Why you removed this test? Is still relevant for reference.

test/Header/ContentLengthTest.php
```diff
@@ -13,56 +13,89 @@
...
-        $header = ContentLength::fromString("Content-Length: xxx\r\n\r\nevilContent");
+        new ContentLength($invalidValue);
     }
 
     /**
-     * @see http://en.wikipedia.org/wiki/HTTP_response_splitting
-     * @group ZF2015-04
+     * @dataProvider invalidValues
      */
-    public function testPreventsCRLFAttackViaConstructor()
```
ezimuel: Again, see comment above.

test/Header/ContentLengthTest.php
```diff
@@ -13,56 +13,89 @@
...
-        $header = new ContentLength("Content-Length: xxx\r\n\r\nevilContent");
+        $header = new ContentLength('123');
+        $header->setFieldValue($invalidValue);
+    }
+
+    /**
+     * Data provider of invalid values
+     * @return array
+     */
+    public function invalidValues()
```
ezimuel: I see what are you doing here and it's fine but we prefer to add only unit tests relevant for the PR.

ezimuel:
@DCMNMarc As I wrote in specific comments this PR should focus only on fixing #26. The other changes that you proposed should be done in a new PR as new feature. Thanks for your contribution!

**Note:** to fix #26 is enough to change this in constructor:
```php
if ($value !== null) {
...
```

ezimuel:
@DCMNMarc I fixed #26 issue with PR #87. Now, if you want you can focus on the new features proposed in this PR. Thanks again!

ezimuel:
Closing this PR in favor of a new one @DCMNMarc.

PR #61, opened 2016-06-15
Improve setOptions behavior

neilime:


ezimuel:
@neilime can you give me more info about it? Do you have any usage example that show this improvement? Thanks!

neilime:
Here is a use case

This is the code of a factory that instanciate an adapter according to a defined configuration: 

```php
// Get http client adapter config
$config = $serviceLocator->get('config');
$httpClientAdapterConfig = $config['http_client_adapter'];
/*
Retrieve an array(
    'type' => '\Zend\Http\Client\Adapter\Curl',
    'options' => array(
        'proxyhost' => 'localhost',  
    ),
);
*/
$adapterType = $httpClientAdapterConfig['type'];
$adapter = new $adapterType();
$adapter->setOptions($httpClientAdapterConfig['options']);
return $adapter;
```
If I want to override this adapter for tests, according to this PR, I just have to update the config for the tests (in the phpunit Bootstrap or something else)
```php
array(
    'http_client_adapter' =>array(
        'type' => '\Zend\Http\Client\Adapter\Test',
        'options' => array(
            'response' => 'HTTP/1.1 200 OK\r\n\r\n',  
        ),
    ),
);
```

I hope my comment is understandable.

ezimuel:
@neilime I'm sorry but I don't see a value for our code base. We already have setter for all the options that you suggested and people can use it without the configuration array that you proposed.

Issue #55, opened 2016-04-15
Incorrect handling of negative MaxAge cookie attribute


milesich:
Currently cookies with negative Max-Age attribute will throw en exception. But according to [RFC6265](http://www.ietf.org/rfc/rfc6265.txt):

>  If the first character of the attribute-value is not a DIGIT or a "-" character, ignore the cookie-av.

I found this problem when I was trying to import a feed using `Zend\Feed\Reader\Reader::import` and in the feed headers was this cookie:

`Set-Cookie: wpSGCacheBypass=0; expires=Fri, 15-Apr-2016 07:10:31 GMT; Max-Age=-3600; path=/`

Xerkus:
Fixed in #120

PR #62, opened 2016-06-19
Removed the negative MaxAge cookie condition

brian978:
Removed the negative MaxAge cookie condition and added check to deny adding cookies with negative MaxAge to Cookies header ref #55

### DIFF
--- a/src/Cookies.php
+++ b/src/Cookies.php
@@ -102,6 +102,10 @@ class Cookies extends Headers
         }
 
         if ($cookie instanceof SetCookie) {
+            if ($cookie->getMaxAge() < 0) {
+                return;
+            }
+
             $domain = $cookie->getDomain();
             $path   = $cookie->getPath();
             if (!isset($this->cookies[$domain])) {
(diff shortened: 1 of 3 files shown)


src/Cookies.php
```diff
@@ -102,6 +102,10 @@ public function addCookie($cookie, $refUri = null)
         }
 
         if ($cookie instanceof SetCookie) {
+            if ($cookie->getMaxAge() < 0) {
```
ezimuel: Actually, according to [RFC 6265](http://www.ietf.org/rfc/rfc6265.txt) we should allow negative number (see also #55 comment).

test/Header/CookieTest.php
```diff
@@ -98,6 +99,14 @@ public function valuesProvider()
         ];
     }
 
+    public function testSetExpiredCookie()
+    {
+        $cookies = new Cookies();
+        $cookies->addCookie("Set-Cookie: wpSGCacheBypass=0; expires=Fri, 15-Apr-2016 07:10:31 GMT; Max-Age=-3600; path=/");
+
+        $this->assertEquals(0, count($cookies->getAllCookies()));
```
ezimuel: As commented above, we should allow negative numbers in Max-Age.

ezimuel:
I'm closing this PR because no answer from @brian978

Issue #84, opened 2016-08-02
Infinite recursion on AbstractAccept header


weckx:
If you create a new Accept (or any other that extends from AbstractAccept) and try to call getFieldValue(), an infinite recursion and a fatal error happens. To reproduce:

```php
$acceptHeader = new Accept();
$acceptHeader->getFieldValue();
```

weckx:
Fixed with PR #85

PR #85, opened 2016-08-02
Fix infinite recursion on AbstractAccept

weckx:
This PR fixes the bug reported in #84  and adds a test for it


test/Header/AcceptTest.php
```diff
@@ -465,4 +465,10 @@ public function testPreventsCRLFAttackViaFromString()
         $this->setExpectedException('Zend\Http\Header\Exception\InvalidArgumentException');
         $header = Accept::fromString("Accept: application/text\r\n\r\nevilContent");
     }
+
+    public function testEmptyValue()
```
ezimuel: I suggest to have a more verbose test name, like `testGetEmptyFieldValue`.

weckx: Sure, I've updated the test name

ezimuel:
@weckx Thanks!

Issue #77, opened 2016-06-28
Http/Header/Cookie Uncaught exception


GeeH:
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7197
User: @Moring
Created On: 2015-02-09T22:51:11Z
Updated At: 2016-03-24T06:58:00Z
Body
PHP Fatal error:  Uncaught exception 'ErrorException' with message 'urlencode() expects parameter 1 to be string, array given' in /home/www/vendor/zendframework/zendframework/library/Zend/Http/Header/Cookie.php:93

referer: http://search.daum.net/

Perhaps add check to $value or graceful catch?

---
### Comment

User: @jaapio
Created On: 2015-02-10T19:15:15Z
Updated At: 2015-02-10T19:15:15Z
Body
I agree with you that this could be nicer, but I'm not sure if you should set any thing else then a scalar value in a Cookie Header. If I understand the rfc about cookies correctly you should use the set-cookie header for this purpose.

---
### Comment

User: @Moring
Created On: 2015-02-10T20:18:59Z
Updated At: 2015-02-10T20:18:59Z
Body
The issue is that the query, when sent to a ZF2 application, triggers an unhandled fatal error.  To me, this can easily become (or is) a penetration test.  It should be handled more gracefully than a fatal error.

---
### Comment

User: @andsol
Created On: 2016-03-24T06:56:34Z
Updated At: 2016-03-24T06:58:00Z
Body
Similar issue, server log full of these errors.

PHP Error: 1: urlencode() expects parameter 1 to be string, array given. File: ...../vendor/zendframework/zend-http/src/Header/Cookie.php, in 101 line.

---

tflori:
open #92 to solve this problem

tflori:
@GeeH I have solved this issue with PR #92 please review it. thx

ezimuel:
@tflori I commented about #92, waiting your reply. Thanks!

Xerkus:
Was this issue fixed by the PR?

tflori:
@icanhazstring can you check if #92 solves this issue?

Moring:
Have not seen this error in the last month of logs, we end to end latest ZF3, so for us, it is fixed.

Xerkus:
thanks

PR #92, opened 2016-10-27
flatten cookies for header

tflori:
Fix issue #77

### DIFF
--- a/src/Header/Cookie.php
+++ b/src/Header/Cookie.php
@@ -97,27 +97,26 @@ class Cookie extends ArrayObject implements HeaderInterface
     {
         $nvPairs = [];
 
-        $cookies = [];
-        $this->flattenCookies($cookies, $this);
-
-        foreach ($cookies as $name => $value) {
+        foreach ($this->flattenCookies($this) as $name => $value) {
             $nvPairs[] = $name . '=' . (($this->encodeValue) ? urlencode($value) : $value);
         }
 
         return implode('; ', $nvPairs);
     }
 
-    protected function flattenCookies(&$result, $data, $prefix = null)
+    protected function flattenCookies($data, $prefix = null)
     {
-        if (!is_array($data) && $this !== $data) {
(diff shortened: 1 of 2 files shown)


jaapio:
Could you provide a unittest for this?

tflori:
@jaapio I created a test can you please recheck?

jaapio:
I can. But I won't be able to approve your pr since I'm not a zf
maintainer.

Op do 8 dec. 2016 13:15 schreef tflori <notifications@github.com>:

tflori:
@weierophinney it would be greate if you could review this pr. thanks in advance.

src/Header/Cookie.php
```diff
@@ -97,13 +97,30 @@ public function getFieldValue()
...
-        foreach ($this as $name => $value) {
+        $cookies = [];
+        foreach ($this->flattenCookies($cookies, $this) as $name => $value) {
             $nvPairs[] = $name . '=' . (($this->encodeValue) ? urlencode($value) : $value);
         }
 
         return implode('; ', $nvPairs);
     }
 
+    protected function flattenCookies(&$result, $data, $prefix = null)
```
ezimuel: Why we need a recursive function here? Do we need to manage array of array values in cookies? I think we should simplify the implementation here, for instance just iterating on keys and values of $value, if it is an array.

ezimuel: Btw, we discourage the usage of passing value by reference in function. A good use case for reference is performance reason, but this is not the case.

tflori: For me the correct way seems to be what php is doing. So yes we need a recursive function. But you are right we can write it without passing by reference.

tflori:
@ezimuel I modified the function not to use pass by reference.

I don't know why coveralls shows a decreased coverage status - the function I added is tested. Nothing else changed.

ezimuel:
@tflori thanks for your contribution! I improved the unit test with a bi-dimensional cookie array.

PR #97, opened 2016-12-02
Add support to sslcafile and sslcapath to cURL adapter

mlocati:
Socket adapter supports the options `sslcafile` and `sslcapath`.

What about supporting them in the cURL Adapter too?

### DIFF
--- a/src/Client/Adapter/Curl.php
+++ b/src/Client/Adapter/Curl.php
@@ -209,6 +209,13 @@ class Curl implements HttpAdapter, StreamInterface
             }
         }
 
+        if (isset($this->config['sslcafile']) && $this->config['sslcafile']) {
+            curl_setopt($this->curl, CURLOPT_CAINFO, $this->config['sslcafile']);
+        }
+        if (isset($this->config['sslcapath']) && $this->config['sslcapath']) {
+            curl_setopt($this->curl, CURLOPT_CAPATH, $this->config['sslcapath']);
+        }
+
         if (isset($this->config['maxredirects'])) {
             // Set Max redirects
             curl_setopt($this->curl, CURLOPT_MAXREDIRS, $this->config['maxredirects']);
--- a/src/Client.php
+++ b/src/Client.php
@@ -110,7 +110,9 @@ class Client implements Stdlib\DispatchableInterface
         'outputstream'    => false,
         'encodecookies'   => true,
         'argseparator'    => null,
-        'rfc3986strict'   => false
+        'rfc3986strict'   => false,
+        'sslcafile'       => null,
+        'sslcapath'       => null,
     ];
 
     /**


src/Client/Adapter/Curl.php
```diff
@@ -209,6 +209,13 @@ public function connect($host, $port = 80, $secure = false)
             }
         }
 
+        if (isset($this->config['sslcafile']) && $this->config['sslcafile']) {
```
ezimuel: I'm thinking that we should check if CURLOPT_SSL_VERIFYPEER is false and in this case throw an Exception if we set `sslcafile` or `sslcapath`. What do you think?

mlocati: Any reason to do that? I would do the opposite: avoid setting the `CURLOPT_CAINFO` and `CURLOPT_CAPATH` curl options if `CURLOPT_SSL_VERIFYPEER` is false (that because I don't know how the underlying libcurl behaves in that case...)

mlocati:
> Please remember to add the unit test 

I'm going to do that

> and send the PR against develop

Done.

> Thanks!

Thanks to you! :wink:

src/Client/Adapter/Curl.php
ezimuel: The idea is to prevent setting CA if CURLOPT_SSL_VERIFYPEER is false. Basically we should allow setting CA only if CURLOPT_SSL_VERIFYPEER is true.
Reading [here](https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html) from the CURL manual, we can use  CURLOPT_CAINFO and CURLOPT_CAPATH only if CURLOPT_SSL_VERIFYPEER is true.

mlocati: That's exactly what I meant.
Instead of throwing an exception (like you suggested), I'd avoid setting `CURLOPT_CAINFO` and `CURLOPT_CAPATH` if `CURLOPT_SSL_VERIFYPEER` is false

ezimuel: I think it's better throwing an exception because we can prevent a potential security issue.
Let me explain, if CURLOPT_SSL_VERIFYPEER is false and someone set the CURLOPT_CAINFO/CAPATH values and we skip these settings, we will have successfully CURL calls without any evidence about the usage of CA. People can assume the usage of CA because of the successfully calls but the CA will be bypassed due to CURLOPT_SSL_VERIFYPEER .
Throwing an exception we can inform people about the correct usage of CA.

mlocati: What about this alternative:
- if `sslverifypeer` is not set and `CURLOPT_CAINFO/CAPATH` are set, we assume that `sslverifypeer` is true
- if `sslverifypeer` is manually set to false, we don't consider `CURLOPT_CAINFO/CAPATH`

ezimuel: CURLOPT_SSL_VERIFYPEER is true by default, so if `sslverifypeer` is not set it's not a problem for CA. The issue comes if `sslverifypeer` is set to false and we set CA. In this case I think we should throw an Exception, instead of a silent action as you proposed. This is a clear misuse of the component,with potential security implications. 
@mlocati why you are against using an Exception for this case?

mlocati: > @mlocati why you are against using an Exception for this case?

@ezimuel Because I love to keep everything as simple as possible: if someone sets `sslverifypeer` to false, he excepts that `CURLOPT_SSL_VERIFYPEER` is set to false, for any value of the other parameters

ezimuel: @mlocati this is not the point, I don't want to change the behaviour of any settings. I would like to prevent wrong settings. Anyway, reading the [CURLOPT_SSL_VERIFYPEER](https://curl.haxx.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html) specification it seems that it works already as is. If it's false, the CA settings are not used anyway.
I think we can just leave the settings as you proposed without any extra check and exceptions. We should only remember to document this behaviour in the documentation.

ezimuel:
@mlocati I added the unit test and the documentation to the PR. Thanks for your contribution!

mlocati:
> @mlocati I added the unit test

@ezimuel Thanks! I wanted to test these configuration keys with some real use case, but I've been very busy lately.

Issue #96, opened 2016-12-02
Split timeout into connectionTimeout and executeTimeout


mlocati:
Currently the adapters support only one timeout option.
BTW, it would be really helpful to have to different timeouts: one used at connection time and another one used after the connection has been made.

This would be really helpful when downloading big files: in that case we currently have to set a big timeout value (for instance 120 seconds). But if the remote server is down, we have to wait 120 seconds before the request time outs.

Having a small connection timeout (eg 5 seconds) and a big execution timeout (eg 120 seconds) would solve this problem.

For instance, for the cURL adapter, we could use `CURLOPT_CONNECTTIMEOUT` and `CURLOPT_TIMEOUT`...

mlocati:
Here's a sample code to implement this for the cURL Adapter: https://github.com/zendframework/zend-http/pull/98

mlocati:
I updated #98 to include the two timeouts in the Socket adapter too

ezimuel:
#98 is merged.

PR #98, opened 2016-12-02
Split timeout in connectTimeout/executeTimeout

mlocati:
See https://github.com/zendframework/zend-http/issues/96

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -103,6 +103,8 @@ class Client implements Stdlib\DispatchableInterface
         'strictredirects' => false,
         'useragent'       => 'Zend\Http\Client',
         'timeout'         => 10,
+        'connecttimeout'  => null,
+        'executetimeout'  => null,
         'adapter'         => 'Zend\Http\Client\Adapter\Socket',
         'httpversion'     => Request::VERSION_11,
         'storeresponse'   => true,
--- a/src/Client.php
+++ b/src/Client.php
@@ -104,7 +104,6 @@ class Client implements Stdlib\DispatchableInterface
         'useragent'       => 'Zend\Http\Client',
         'timeout'         => 10,
         'connecttimeout'  => null,
-        'executetimeout'  => null,
         'adapter'         => 'Zend\Http\Client\Adapter\Socket',
         'httpversion'     => Request::VERSION_11,
         'storeresponse'   => true,
--- a/src/Client/Adapter/Curl.php
+++ b/src/Client/Adapter/Curl.php
@@ -195,17 +195,33 @@ class Curl implements HttpAdapter, StreamInterface
             curl_setopt($this->curl, CURLOPT_PORT, intval($port));
         }
 
-        if (isset($this->config['timeout'])) {
+        if (isset($this->config['connecttimeout'])) {
+            $connectTimeout = $this->config['connecttimeout'];
+        } elseif (isset($this->config['timeout'])) {
+            $connectTimeout = $this->config['timeout'];
+        } else {
+            $connectTimeout = null;
+        }
+        if ($connectTimeout !== null) {
             if (defined('CURLOPT_CONNECTTIMEOUT_MS')) {
-                curl_setopt($this->curl, CURLOPT_CONNECTTIMEOUT_MS, $this->config['timeout'] * 1000);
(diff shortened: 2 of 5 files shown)


src/Client/Adapter/Curl.php
```diff
@@ -195,17 +195,33 @@ public function connect($host, $port = 80, $secure = false)
...
             if (defined('CURLOPT_CONNECTTIMEOUT_MS')) {
-                curl_setopt($this->curl, CURLOPT_CONNECTTIMEOUT_MS, $this->config['timeout'] * 1000);
+                curl_setopt($this->curl, CURLOPT_CONNECTTIMEOUT_MS, $connectTimeout * 1000);
             } else {
-                curl_setopt($this->curl, CURLOPT_CONNECTTIMEOUT, $this->config['timeout']);
+                curl_setopt($this->curl, CURLOPT_CONNECTTIMEOUT, $connectTimeout);
             }
+        }
 
+        if (isset($this->config['executetimeout'])) {
```
ezimuel: Instead of add a new parameter `executetimeout` why we don't use the old one `timeout`? The meaning is the same and we can be aligned with the curl constant naming.

src/Client/Adapter/Socket.php
```diff
@@ -240,12 +240,19 @@ public function connect($host, $port = 80, $secure = false)
                 $flags |= STREAM_CLIENT_PERSISTENT;
             }
 
+            if (isset($this->config['connecttimeout'])) {
+                $connectTimeout = $this->config['connecttimeout'];
+            } elseif (isset($this->config['timeout'])) {
+                $connectTimeout = $this->config['timeout'];
+            } else {
+                $connectTimeout = @ini_get('default_socket_timeout');
```
ezimuel: We generally don't use the `@` operator. Anyway, in this case the value `default_socket_timeout` is available since PHP 4.3.0.

src/Client/Adapter/Socket.php
```diff
@@ -597,8 +612,16 @@ protected function _checkSocketReadTimeout()
             $timedout = $info['timed_out'];
             if ($timedout) {
                 $this->close();
+                if (isset($this->config['executetimeout'])) {
```
ezimuel: As commented for the curl adapter, even in this case we should continue to use the `timeout` parameter instead of adding a new one (`executetimeout`).

src/Client/Adapter/Socket.php
```diff
@@ -266,7 +273,15 @@ public function connect($host, $port = 80, $secure = false)
             }
 
             // Set the stream timeout
-            if (!stream_set_timeout($this->socket, (int) $this->config['timeout'])) {
+            if (isset($this->config['executetimeout'])) {
+                $executeTimeout = $this->config['executetimeout'];
+            } elseif (isset($this->config['timeout'])) {
+                $executeTimeout = $this->config['timeout'];
+            } else {
+                $executeTimeout = @ini_get('default_socket_timeout');
```
ezimuel: Same comment of curl adapter. Don't use the `@` operator.

src/Client.php
```diff
@@ -103,6 +103,8 @@ class Client implements Stdlib\DispatchableInterface
         'strictredirects' => false,
         'useragent'       => 'Zend\Http\Client',
         'timeout'         => 10,
+        'connecttimeout'  => null,
+        'executetimeout'  => null,
```
ezimuel: We should use the old value `timeout` instead of adding `executetimeout`.

mlocati:
> @mlocati I did a first review of your PR. As commented in details, I think we can use the old parameter timeout instead of executetimeout.

Done in https://github.com/zendframework/zend-http/pull/98/commits/6da4e3dc5fcfca7636247a44bb451b6e97b3d149

> The PR needs to be completed with unit test to cover the new features, can you add it?

I'm going to do this.

> Moreover, the PR should be send to develop because this is a new feature.
Thanks!

PR updated: now it's against develop

mlocati:
> The PR needs to be completed with unit test to cover the new features, can you add it?

Done in https://github.com/zendframework/zend-http/pull/98/commits/0bff41b0c09403473ac78ee3e53fb736da2bf0f7

ezimuel:
@mlocati thanks for your contribution!

PR #106, opened 2017-01-24
Fix coding style

mlocati:


ezimuel:
@mlocati thanks for this fix. I was doing it but you were faster :)

PR #107, opened 2017-01-25
Allow the usage of "Expires: 0" header

### LINKED ISSUE #103: AbstractDate::fromString('Expires: 0') fails
Hi, we developed a custom Zend Event Listener for our Apigility application, this listener adds custom HTTP headers using the `MvcEvent::EVENT_ROUTE` Zend event.

In our response headers we send an `Expires: 0` header, this is to prevent bugs with Internet Explorer which seems to cache our payloads if we do not explicitly provide HTTP headers to disable caching.

When we provided the `Expires: 0` HTTP header Zend Http fails with the following exception.

```

Fatal error: Uncaught Exception: DateTime::__construct(): Failed to parse time string (0) at position 0 (0): Unexpected character in D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Header\AbstractDate.php:167 Stack trace: 
#0 D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Header\AbstractDate.php(167): DateTime->__construct('0', Object(DateTimeZone)) 
#1 D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Header\AbstractDate.php(86): Zend\Http\Header\AbstractDate->setDate('0') 
#2 D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Headers.php(448): Zend\Http\Header\AbstractDate::fromString('Expires: 0') 
#3 D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Headers.php(352): Zend\Http\Headers->lazyLoadHeader(8) 
#4 D:\workspace\vgm-web-services\vendor\zendframework\zend-mvc\src\ResponseSender\AbstractResponseSender.php(30): Zend\Http\Headers->current() 
#5 D:\workspace\vgm-web-services\vendor\zendframework\zend-mvc\src\ResponseSender\PhpEnvironmentRespon in D:\workspace\vgm-web-services\vendor\zendframework\zend-http\src\Header\AbstractDate.php on line 169
```

As you can see the problem comes from the following instruction.

```
Zend\Http\Header\AbstractDate::fromString('Expires: 0')
```

IMO this behavior is not normal because as explained in the RFC https://tools.ietf.org/html/rfc2616#section-14.21 values which are not compliant with the date format must be treated. 

> HTTP/1.1 clients and caches MUST treat other invalid date formats,
   especially including the value "0", as in the past (i.e., "already
   expired").

So I think Zend should consider that invalid dates are equivalent to `Thu, 01 Jan 1970 00:00:00 GMT`. 

Do your agree with that ? 

Perhaps this issue is also linked to https://github.com/zendframework/zend-http/issues/9.

Thanks

ezimuel:
This PR fixes #103, allowing `Expires: 0` header. I used the timestamp 0 that is equivalent to "Thu, 01 Jan 1970 00:00:00 GMT".

### DIFF
--- a/src/Header/Expires.php
+++ b/src/Header/Expires.php
@@ -32,7 +32,7 @@ class Expires extends AbstractDate
 
     public function setDate($date)
     {
-        if ($date === '0') {
+        if ($date === '0' || $date === 0) {
             $date = date(DATE_W3C, 0); // Thu, 01 Jan 1970 00:00:00 GMT
         }
         return parent::setDate($date);
--- a/src/Header/Expires.php
+++ b/src/Header/Expires.php
@@ -9,6 +9,9 @@
 
 namespace Zend\Http\Header;
 
+use DateTime;
+use DateTimeZone;
+
 /**
  * Expires Header
  *
@@ -24,5 +27,14 @@ class Expires extends AbstractDate
     public function getFieldName()
     {
         return 'Expires';
+    }
+
+
+    public function setDate($date)
+    {
+        if ($date === '0') {
+            $date = date(DATE_W3C, 0); // Thu, 01 Jan 1970 00:00:00 GMT
+        }
+        return parent::setDate($date);
     }
 }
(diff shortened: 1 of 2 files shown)


src/Header/Expires.php
```diff
@@ -25,4 +28,13 @@ public function getFieldName()
     {
         return 'Expires';
     }
+
+
+    public function setDate($date)
+    {
+        if ($date === '0') {
```
weierophinney: Should this also allow an integer 0 value?

ezimuel: I think this is reasonable. Just update the PR.

PR #109, opened 2017-02-01
Add to Proxy adapter all the options of the Socket adapter

mlocati:
`sslcafile` and `sslverifypeername` were present in the `Socket` adapter but not in the `Proxy` adapter.

We need all the options in order to avoid PHP warnings about accessing invalid array keys.

### DIFF
--- a/src/Client/Adapter/Proxy.php
+++ b/src/Client/Adapter/Proxy.php
@@ -31,19 +31,21 @@ class Proxy extends Socket
      * @var array
      */
     protected $config = [
+        'persistent'         => false,
         'ssltransport'       => 'ssl',
         'sslcert'            => null,
         'sslpassphrase'      => null,
         'sslverifypeer'      => true,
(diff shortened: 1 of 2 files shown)


Xerkus:
@mlocati please add test that ensures all of the socket config options are present in proxy

mlocati:
@Xerkus done.

Xerkus:
Do you really need those proxies? getConfig() could work just as well

mlocati:
@Xerkus Of course you're right... I looked at the `Proxy` methods only and I didn't found a way to get the `$config` property. But obviously the `getConfig` method of the `Socket` class works fine.
PR updated.

Xerkus:
looks good, thanks

Xerkus:
one last thing, rebase it on master please. Patch level change

mlocati:
> rebase it on master

@Xerkus done.

PR #112, opened 2017-02-07
Very inefficient decoding of large chunked messages!

domoran:
In one of our legacy projects we download a ~200MB file in memory using chunked encoding. Calling getBody() on this (which calls decodeChunkedBody()) will take > 5 minutes for processing the response() since creates very inefficient copies of the body with each chunk (substr). 

To make this more efficient we can use the offset parameter, to avoid copying the body during iteration. 

Tests still passed.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -505,17 +505,25 @@ class Response extends AbstractMessage implements ResponseInterface
     {
         $decBody = '';
 
-        while (trim($body)) {
-            if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m)) {
-                throw new Exception\RuntimeException(
-                    "Error parsing body - doesn't seem to be a chunked message"
-                );
+        $offset = 0;
+        
+        while (true) {
+            if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m, 0, $offset)) {
+                if (!empty(trim(substr($body, $offset)))) {
+                    // Message was not consumed completely! 
+                    throw new Exception\RuntimeException(
+                            "Error parsing body - doesn't seem to be a chunked message"
+                    );
+                } else {
+                    // Message was consumed completely
+                    break; 
+                }
             }
 
             $length   = hexdec(trim($m[1]));
             $cut      = strlen($m[0]);
-            $decBody .= substr($body, $cut, $length);
-            $body     = substr($body, $cut + $length + 2);
+            $decBody .= substr($body, $offset+$cut, $length);
+            $offset += $cut + $length + 2;
         }
 
         return $decBody;
--- a/src/Response.php
+++ b/src/Response.php
@@ -506,23 +506,23 @@ class Response extends AbstractMessage implements ResponseInterface
         $decBody = '';
 
         $offset = 0;
-        
+
         while (true) {
             if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m, 0, $offset)) {
-                if (!empty(trim(substr($body, $offset)))) {
-                    // Message was not consumed completely! 
+                if (! empty(trim(substr($body, $offset)))) {
+                    // Message was not consumed completely!
                     throw new Exception\RuntimeException(
-                            "Error parsing body - doesn't seem to be a chunked message"
+                        "Error parsing body - doesn't seem to be a chunked message"
                     );
                 } else {
                     // Message was consumed completely
-                    break; 
+                    break;
                 }
             }
 
             $length   = hexdec(trim($m[1]));
             $cut      = strlen($m[0]);
-            $decBody .= substr($body, $offset+$cut, $length);
+            $decBody .= substr($body, $offset + $cut, $length);
             $offset += $cut + $length + 2;
         }
 
--- a/src/Response.php
+++ b/src/Response.php
@@ -509,15 +509,14 @@ class Response extends AbstractMessage implements ResponseInterface
 
         while (true) {
             if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m, 0, $offset)) {
-                if (! empty(trim(substr($body, $offset)))) {
+                if (trim(substr($body, $offset))) {
                     // Message was not consumed completely!
                     throw new Exception\RuntimeException(
-                        "Error parsing body - doesn't seem to be a chunked message"
+                        'Error parsing body - doesn\'t seem to be a chunked message'
                     );
-                } else {
-                    // Message was consumed completely
-                    break;
                 }
+                // Message was consumed completely
+                break;
             }
 
             $length   = hexdec(trim($m[1]));
--- a/test/ResponseTest.php
+++ b/test/ResponseTest.php
@@ -13,6 +13,7 @@ use Zend\Http\Exception\RuntimeException;
 use Zend\Http\Header\GenericHeader;
 use Zend\Http\Headers;
 use Zend\Http\Response;
+use Zend\Http\Headers;
 
 class ResponseTest extends TestCase
 {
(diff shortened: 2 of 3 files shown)


Ocramius:
Requires a performance test to prevent regressions. Unsure how to achieve that besides using relative timing of the regex applied to some data, and a large chunk of text passed to this method.

domoran:
I added a performance test, by crafting a "worst case" packet and measuring relative timing. 

Verified that test fails on old version: 
git checkout e89dac73d56b18023b8e2e41074e64bb3abbdfda src/Response.php

src/Response.php
```diff
@@ -495,17 +495,25 @@ protected function decodeChunkedBody($body)
...
-        while (trim($body)) {
-            if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m)) {
-                throw new Exception\RuntimeException(
-                    "Error parsing body - doesn't seem to be a chunked message"
-                );
+        $offset = 0;
+
+        while (true) {
+            if (! preg_match("/^([\da-fA-F]+)[^\r\n]*\r\n/sm", $body, $m, 0, $offset)) {
+                if (! empty(trim(substr($body, $offset)))) {
```
webimpress: `! empty` seems to be useless here

src/Response.php
```diff
@@ -495,17 +495,25 @@ protected function decodeChunkedBody($body)
...
+                    // Message was not consumed completely!
+                    throw new Exception\RuntimeException(
+                        "Error parsing body - doesn't seem to be a chunked message"
```
webimpress: Can we use single quotes around the error message string above?

src/Response.php
```diff
@@ -495,17 +495,25 @@ protected function decodeChunkedBody($body)
...
+                    );
+                } else {
```
webimpress: I think we don't need `else` statement here.

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
         $this->assertEquals('c0cc9d44790fa2a58078059bab1902a9', md5($res->getContent()));
     }
 
+    /**
+     * @param number $chunksize the data size of the chunk to create
```
webimpress: Please use `int` param type instead of `number`

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+     */
+    private function makeChunk($chunksize)
+    {
+        $chunkdata = str_repeat("W", $chunksize);
+        return "$chunksize\r\n$chunkdata\r\n";
+    }
+
+    /**
+     * @param Response $response
+     * @return the time that calling the getBody function took on the response
```
webimpress: Please add return type.

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
         $this->assertEquals('c0cc9d44790fa2a58078059bab1902a9', md5($res->getContent()));
     }
 
+    /**
+     * @param number $chunksize the data size of the chunk to create
+     * @return string a chunk of data for embedding inside a chunked response
+     */
+    private function makeChunk($chunksize)
+    {
+        $chunkdata = str_repeat("W", $chunksize);
```
webimpress: Please use single quotes `'` above instead of double `"`.

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+        return "$chunksize\r\n$chunkdata\r\n";
```
webimpress: Not sure here, but maybe `sprintf` and/or `PHP_EOL`?

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+    }
+
+    /**
+     * @param Response $response
+     * @return the time that calling the getBody function took on the response
+     */
+    private function getTimeForGetBody(Response $response)
+    {
+        $time_start = microtime(true);
```
webimpress: Please use camelCase variable names.

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+        $response->setContent($responseData);
+        $time1 = $this->getTimeForGetBody($response);
+
+        // Get baseline for timing: 1000 x 1 Byte chunks
+        $responseData2 = $responseData . $this->makeChunk(1000000);
+        $response->setContent($responseData2);
+        $time2 = $this->getTimeForGetBody($response);
+
+        // Make sure that the worst case packet will have an equal timing as the baseline
+        $errMsg = "Chunked response is not parsing large packets efficiently: " . ($time2 / $time1);
```
webimpress: Please use `sprintf` there and single quotes.

test/ResponseTest.php
```diff
@@ -176,6 +177,54 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+        $this->assertTrue(2 > ($time2 / $time1), $errMsg);
```
webimpress: Maybe we can use here `assertLessThan`?
Brackets around `$time2 / $time1` are useless.

test/ResponseTest.php
```diff
@@ -10,6 +10,7 @@
 namespace ZendTest\Http;
 
 use Zend\Http\Response;
+use Zend\Http\Headers;
```
webimpress: Please have all imports in alphabetical order.

domoran:
Fixed formatting comments.

domoran:
Hmm ... the test seems flaky, failed under PHP 7 for the last build ... I cannot think of a way to make the test more stable except by increasing its runtime.  (More data, increase ratio) Any ideas?

Ocramius:
@domoran flakiness can probably be avoided by repeating the test, but indeed, it's a time-based test, so it will always be a risky one. Make sure the first iteration is removed from measurement to avoid autolading and other possible performance issues there.

Ocramius:
Sorry for the previous commits - previous review from me was never submitted.

test/ResponseTest.php
```diff
@@ -176,6 +177,60 @@ public function testChunkedResponseCaseInsensitiveZF5438()
...
+     */
+    private function getTimeForGetBody(Response $response)
+    {
+        $timeStart = microtime(true);
+        $response->getBody();
+        return microtime(true) - $timeStart;
+    }
+
+    /**
+     * @small
```
Ocramius: What's `@small` here?

Ocramius: Does the test fail with the previous code? (just asking, since can't try it on my machine atm)

test/ResponseTest.php
```diff
@@ -176,6 +177,60 @@ public function testChunkedResponseCaseInsensitiveZF5438()
         $this->assertEquals('c0cc9d44790fa2a58078059bab1902a9', md5($res->getContent()));
     }
 
+    /**
+     * @param int $chunksize the data size of the chunk to create
+     * @return string a chunk of data for embedding inside a chunked response
+     */
+    private function makeChunk($chunksize)
+    {
+        $chunkdata = str_repeat('W', $chunksize);
```
Ocramius: Can be inlined

domoran:
I dont get it. There must be a logic error in the test or a problem on php7. 

How can one more regexp iteration inside the decodeChunkedBody be factor [30-50(!)](https://travis-ci.org/zendframework/zend-http/jobs/199619906) slower than the baseline? This happens only on travis-ci and only with php 7 ... 

On my machine I cannot reproduce. Even with php7 I get a ratio of 3-5 ... But with php5.6 there is the expected runtime ratio of 1 ... Maybe an issue with the regexp implementation on PHP 7?

Oh boy. Committing to Zend Framework is hard work.

test/ResponseTest.php
domoran: On my machine the test does not fail at all. Neither with php7 nor php 5.6 ... small will be removed.

Ocramius:
> Oh boy. Committing to Zend Framework is hard work.

Nah, you just found an annoyingly hard edge case.

Did you check whether you are running with xdebug or not? Try running tests with `php -n` (will probably fail on some curl-based scenarios).

If we can help out, let us know, we can certainly test on other environments as well.

PHP 7 has indeed changed many things in the regex engine, btw.

test/ResponseTest.php
Ocramius: @domoran I was asking if the test fails when removing the patch to the chunk decoding block.

domoran: Yes, without the path the ratios are on my machine 2000-5000 under PHP7 and about 250 on PHP5.6 ... Test runs > 60 secs.  

I can increase the ratio here to make the test pass, but there must be some other problem. I would not expect a ratio much larger than 1, after all this is only one iteration more than before.

Ocramius: Alright, going up with those timings is not gonna help, so I suppose that a different approach needs to be taken.

Could you poke me on Friday? I can probably work on it while traveling.

weierophinney:
Thanks, @domoron! We have updated the test suite recently, and those updates ensured your performance-based tests now pass reliably! Merged for next release.

PR #115, opened 2017-03-03
Updated dependencies - PHPUnit 5.7 and 6.0, dropped PHP5.5 support, QA Tools

webimpress:
- PHPUnit ^6.0 || ^5.7 - Updated all tests to support both versions.

- Added docheader tool to checks all license headers and update headers in all files.

- Updated composer scripts and CONTRIBUTING to refer these scripts.

- Updated travis configuration to run tests with lowest/locked/latest dependencies on PHP 5.6/7.0/7.1.

- Dropped PHP 5.5 support.

- Many CS fixes (fixed indents, array trailing commas, double quotes, formatting)


froschdesign:
@webimpress 

Thanks for your work, but:

> - PHPUnit ^6.0 || ^5.7
> - Updated all tests to support both versions.
> - Added docheader tool to checks all license headers and update headers in all files.
> - Updated composer scripts and CONTRIBUTING to refer these scripts.
> - Updated travis configuration to run tests with lowest/locked/latest dependencies on PHP 5.6/7.0/7.1.
> - Dropped PHP 5.5 support.

- changes in 214 files
- 3.228 new lines
- 1.377 removed lines
- completely different topics

All this in one commit?!

Btw. This is more than a ugly job for a reviewer.

webimpress:
@froschdesign Yeah... I know it's a lot for one commit. I can split it for just update PHPUnit and then license headers, because all of it is mainly because of that.

The problem is also with docheader that we can't exclude some files. Maybe we should exclude the whole `test` directory? Do you want me to update the PR and split it into smaller commits?

webimpress:
Ok, I'll split it later into smaller commits and push force. Sorry for huge commit ...

froschdesign:
> The problem is also with docheader that we can't exclude some files.

I mean this: malukenho/docheader#13

webimpress:
@froschdesign PR updated. Please have a look and let me know what you think. Thanks!

src/Client.php
```diff
@@ -1264,14 +1275,15 @@ protected function prepareBody()
...
+                $body .= '--' . $boundary . '--' . "\r\n";
             } elseif (stripos($this->getEncType(), self::ENC_URLENCODED) === 0) {
                 // Encode body as application/x-www-form-urlencoded
                 $body = http_build_query($this->getRequest()->getPost()->toArray(), null, '&');
             } else {
-                throw new Client\Exception\RuntimeException(
-                    "Cannot handle content type '{$this->encType}' automatically"
-                );
+                throw new Client\Exception\RuntimeException(sprintf(
+                    'Cannot handle content type "%s" automatically',
```
Xerkus: Those swapped quotes broken tests. Is this change required by CS tool?

webimpress:
@Xerkus reverted. All green now. Thanks!

composer.json
```diff
@@ -14,16 +14,17 @@
         }
     },
     "require": {
-        "php": "^5.5 || ^7.0",
-        "zendframework/zend-loader": "^2.5",
-        "zendframework/zend-stdlib": "^2.5 || ^3.0",
-        "zendframework/zend-uri": "^2.5",
-        "zendframework/zend-validator": "^2.5"
+        "php": "^5.6 || ^7.0",
+        "zendframework/zend-loader": "^2.5.1",
```
Xerkus: why did you bump those dependencies?

webimpress: Why you think it should be outside the scope of this PR? IMHO it should be in this scope, because I've also added testing with lowest/locked/latest dependencies. Usually we bump to the latest version (which contain all updates - and previous version was not LTS) to force users to use versions with all fixes. I think we should then have these dependencies bump here. Otherwise we can find many things to split this PR into multiple smaller PRs, but what is the point?

Xerkus: Updating minimum versions for those packages is unrelated to either of 1) phpunit 2) php version bump 3) coding style fixes 4) travis, unless they were failing  with new test setup.

Hence it is outside of the scope of this PR.

Anyway, disregard that.

Issue #116, opened 2017-03-09
Headers invalid argument in addHeaderLine method


necromant2005:
Headers failing by sending empty "user-agent"
The main issue in Headers::addHeaders
```
if (is_string($value)) {
   $this->addHeaderLine($value);
```
But below in method addHeaderLine signature
```
    public function addHeaderLine($headerFieldNameOrLine, $fieldValue = null)
    {
        $matches = null;
        if (preg_match('/^(?P<name>[^()><@,;:\"\\/\[\]?=}{ \t]+):.*$/', $headerFieldNameOrLine, $matches)
            && $fieldValue === null) {
            // is a header
            $headerName = $matches['name'];
            $headerKey  = static::createKey($matches['name']);
            $line = $headerFieldNameOrLine;
        } elseif ($fieldValue === null) {
            throw new Exception\InvalidArgumentException('A field name was provided without a field value');
        }
```

Witch is wrong typing NULL on string only fields

weierophinney:
@necromant2005 —

I'm unclear what you expect to occur. If the idea is to unset the header, you would need to use:

```php
if ($headers->has($header)) {
    $headers->removeHeader($headers->get($name));
}
```

If the idea is to set an empty value, as @Xerkus notes, that's invalid per the HTTP specifications. You *could* do something like this:

```php
$headers->addHeaderLine($header, '');
```

Both of the above would require that you _validate_ the value being provided before you try and create the header, which is likely a good idea anyways.

We do not intend to support the usage `$headers->addHeaderLine($value)` where the value is not a fully-formed header line, but just the header name, as its purpose would be too vague:

- Should it remove such a header, if it's already present?
- Or set a header with an empty string as a value (invalid by the specification)?

Neither feels like a valid option.

necromant2005:
I'm getting error on production because some bots comes with empty "user-agent" headers
```
file:vendor/zendframework/zend-servicemanager/src/ServiceManager.php | 
ln:765 | 
msg: message => 'Uncaught Zend\\Http\\Exception\\InvalidArgumentException: A field name was provided without a field value in vendor/zendframework/zend-http/src/Headers.php:192

Stack trace:
#0 vendor/zendframework/zend-http/src/Headers.php(155): Zend\\Http\\Headers->addHeaderLine(\'User-Agent\')
#1 vendor/zendframework/zend-http/src/PhpEnvironment/Request.php(238): Zend\\Http\\Headers->addHeaders(Array)
[...]

PR #117, opened 2017-03-09
fix header missed value exception

necromant2005:
Headers failing by sending empty "user-agent"
The main issue in Headers::addHeaders
```
if (is_string($value)) {
   $this->addHeaderLine($value);
```
But below in method addHeaderLine signature
```
    public function addHeaderLine($headerFieldNameOrLine, $fieldValue = null)
    {
        $matches = null;
        if (preg_match('/^(?P<name>[^()><@,;:\"\\/\[\]?=}{ \t]+):.*$/', $headerFieldNameOrLine, $matches)
            && $fieldValue === null) {
            // is a header
            $headerName = $matches['name'];
            $headerKey  = static::createKey($matches['name']);
            $line = $headerFieldNameOrLine;
        } elseif ($fieldValue === null) {
            throw new Exception\InvalidArgumentException('A field name was provided without a field value');
        }
```

Witch is wrong typing NULL on string only fields

### DIFF
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -179,17 +179,14 @@ class Headers implements Countable, Iterator
      * @param string $fieldValue optional
      * @return Headers
      */
-    public function addHeaderLine($headerFieldNameOrLine, $fieldValue = null)
+    public function addHeaderLine($headerFieldNameOrLine, $fieldValue = '')
     {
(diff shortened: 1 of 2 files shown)


Xerkus:
As noted in #116 no value headers are invalid

PR #121, opened 2017-04-12
cast to (int) for Curl adapter timeout config

samsonasik:
same as what applied into Socket adapter

### DIFF
--- a/src/Client/Adapter/Curl.php
+++ b/src/Client/Adapter/Curl.php
@@ -202,9 +202,9 @@ class Curl implements HttpAdapter, StreamInterface
         }
 
         if (isset($this->config['connecttimeout'])) {
-            $connectTimeout = $this->config['connecttimeout'];
+            $connectTimeout = (int) $this->config['connecttimeout'];
         } elseif (isset($this->config['timeout'])) {
-            $connectTimeout = $this->config['timeout'];
+            $connectTimeout = (int) $this->config['timeout'];
         } else {
             $connectTimeout = null;
         }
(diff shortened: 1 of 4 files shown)


Xerkus:
What is the reason for accepting non-integer timeout values?
I would say exception should be thrown instead.

samsonasik:
just for consistent with socket adapter. if it should be get exception, the in socket adapter should get exception as well?

Xerkus:
That would be my preference, yes. I will need to look into it sometime later, may be i am not seeing something.

src/Client/Adapter/Curl.php
```diff
@@ -203,9 +203,9 @@ public function connect($host, $port = 80, $secure = false)
         }
 
         if (isset($this->config['connecttimeout'])) {
-            $connectTimeout = $this->config['connecttimeout'];
+            $connectTimeout = (int) $this->config['connecttimeout'];
```
weierophinney: I would prefer the following approach:

- Grab the config value.
- If it is not an integer or not a numeric string, throw an `InvalidArgumentException`.
- Cast to integer.

That approach ensures strings like `timeout` do not get cast to `0`, which is obviously incorrect.

samsonasik: do we need to check `$this->config['timeout']` and $this->config['connecttimeout'] in `setOptions()` function ? and change it at Socket adapter as well?

samsonasik: I've updated with integer and numeric check, applied to socket adapter as well.

weierophinney:
Thanks, @samsonasik!

PR #122, opened 2017-04-13
Avoid opening output stream twice

mlocati:
When we set the `outputstream` option, the output file gets opened twice: [here](https://github.com/zendframework/zend-http/blob/8c23aaae03d9dcb0f2bdb7d3372cc306f411d537/src/Client.php#L722) and [here](https://github.com/zendframework/zend-http/blob/8c23aaae03d9dcb0f2bdb7d3372cc306f411d537/src/Client.php#L938).

The second handle can be managed (eg closed) by accessing the `Zend\Http\Response\Stream\getStream()` method, but the first handle is simply lost: the file remains open (ie locked) until the end of the execution of the PHP script.

Let's reuse the first file handle.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -69,6 +69,11 @@ class Client implements Stdlib\DispatchableInterface
     protected $streamName = null;
 
     /**
+     * @var resource|null
+     */
+    protected $streamHandle = null;
+
+    /**
      * @var array of Header\SetCookie
      */
     protected $cookies = [];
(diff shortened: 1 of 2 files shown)


mlocati:
I updated this PR by adding a test to check that we can delete the downloaded file (previously [this assertion](https://github.com/zendframework/zend-http/pull/122/commits/45ea0f05f9e3954d21adf314d1edec034075fd59#diff-787cc261f147de81449651cc0ca10e45R532) would have failed on Windows).

src/Client.php
```diff
@@ -918,11 +923,17 @@ public function send(Request $request = null)
                 throw new Client\Exception\RuntimeException('Adapter does not support streaming');
             }
 
+            $this->streamHandle = null;
```
Xerkus: I don't like that this code is relying on side effects

mlocati: Do you have a better suggestion?

weierophinney:
Thanks, @mlocati.

PR #125, opened 2017-08-09
 replace count() usage to empty() when possible

samsonasik:
ref: http://maettig.com/code/php/php-performance-benchmarks.php#empty

PR #129, opened 2017-08-24
Small optimalizations with count and strlen

snapshotpl:
Something more to @samsonasik pr in #125

### DIFF
--- a/src/Header/AbstractAccept.php
+++ b/src/Header/AbstractAccept.php
@@ -170,7 +170,7 @@ abstract class AbstractAccept implements HeaderInterface
             foreach ($paramsStrings as $param) {
                 $explode = explode('=', $param, 2);
 
-                if (count($explode) === 2) {
+                if (isset($explode[1])) {
                     $value = trim($explode[1]);
                 } else {
                     $value = null;
@@ -349,7 +349,7 @@ abstract class AbstractAccept implements HeaderInterface
                         $pieces
                     );
 
-                    if (count($pieces) == 3 &&
+                    if (count($pieces) === 3 &&
                         (version_compare($pieces[1], $match1->params[$key], '<=')  xor
                          version_compare($pieces[2], $match1->params[$key], '>=')
                         )
@@ -438,10 +438,12 @@ abstract class AbstractAccept implements HeaderInterface
             //@todo count number of dots in case of type==application in subtype
 
             // So far they're still the same. Longest string length may be more specific
-            if (strlen($a->raw) == strlen($b->raw)) {
+            $aLength = strlen($a->raw);
+            $bLength = strlen($b->raw);
+            if ($aLength === $bLength) {
                 return 0;
             }
-            return (strlen($a->raw) > strlen($b->raw)) ? -1 : 1;
+            return ($aLength > $bLength) ? -1 : 1;
         };
 
         usort($this->fieldValueParts, $sort);
--- a/src/Header/ContentType.php
+++ b/src/Header/ContentType.php
@@ -54,7 +54,7 @@ class ContentType implements HeaderInterface
         $mediaType         = array_shift($parts);
         $header = new static($value, trim($mediaType));
 
-        if (!empty($parts)) {
+        if (! empty($parts)) {
             $parameters = [];
             foreach ($parts as $parameter) {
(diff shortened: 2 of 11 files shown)


src/Header/ContentType.php
```diff
@@ -54,7 +54,7 @@ public static function fromString($headerLine)
         $mediaType         = array_shift($parts);
         $header = new static($value, trim($mediaType));
 
-        if (count($parts) > 0) {
+        if (!empty($parts)) {
```
samsonasik: space after not operator

src/Header/AbstractAccept.php
```diff
@@ -170,7 +170,7 @@ protected function getParametersFromFieldValuePart($fieldValuePart)
             foreach ($paramsStrings as $param) {
                 $explode = explode('=', $param, 2);
 
-                if (count($explode) === 2) {
+                if (isset($explode[1])) {
```
Ocramius: Should check that key 0 exists, and key 2 doesn't.

src/Header/AbstractAccept.php
```diff
@@ -438,10 +438,12 @@ protected function sortFieldValueParts()
             //@todo count number of dots in case of type==application in subtype
 
             // So far they're still the same. Longest string length may be more specific
-            if (strlen($a->raw) == strlen($b->raw)) {
+            $aLength = strlen($a->raw);
+            $bLength = strlen($b->raw);
+            if ($aLength === $bLength) {
```
Ocramius: Spaceship operator: only case where I ever suggested its usage.

src/Header/ContentType.php
```diff
@@ -54,7 +54,7 @@ public static function fromString($headerLine)
         $mediaType         = array_shift($parts);
         $header = new static($value, trim($mediaType));
 
-        if (count($parts) > 0) {
+        if (! empty($parts)) {
             $parameters = [];
```
Ocramius: `empty()` not needed: the `! ` is sufficient

src/Header/Cookie.php
```diff
@@ -60,7 +60,7 @@ public static function fromString($headerLine)
         $arrayInfo = [];
         foreach ($nvPairs as $nvPair) {
             $parts = explode('=', $nvPair, 2);
-            if (count($parts) != 2) {
+            if (! isset($parts[1])) {
```
Ocramius: Should check for key 2. Test needed

Ocramius: Never mind, explode is limited to 2 elements

src/Headers.php
```diff
@@ -153,12 +153,22 @@ public function addHeaders($headers)
...
                 if (is_string($value)) {
                     $this->addHeaderLine($value);
-                } elseif (is_array($value) && count($value) == 1) {
-                    $this->addHeaderLine(key($value), current($value));
-                } elseif (is_array($value) && count($value) == 2) {
-                    $this->addHeaderLine($value[0], $value[1]);
-                } elseif ($value instanceof Header\HeaderInterface) {
+                    continue;
+                }
+                if (is_array($value)) {
```
Ocramius: Move contents of this block to a private method

src/PhpEnvironment/Request.php
```diff
@@ -95,7 +95,7 @@ public function getContent()
     {
         if (empty($this->content)) {
             $requestBody = file_get_contents('php://input');
-            if (strlen($requestBody) > 0) {
+            if (! empty($requestBody)) {
```
Ocramius: `empty()` not needed: `!` is sufficient

samsonasik: for `bool` comparison usage, with `if ($requestBody) {` should be enough

src/Header/AbstractAccept.php
weierophinney: Can't use it here; this component still supports PHP 5.6.

src/Header/AbstractAccept.php
weierophinney: The explode operation was given a limit of 2, so there's no chance a key 2 will ever exist.

weierophinney:
Closing, as the author has not incorporated any of the requested changes.

PR #131, opened 2017-10-10
Handle 100 continue responses proposal

t1gor:
Attempt to fix a `100 Continue` responses issue.

### DIFF
--- a/src/Response.php
+++ b/src/Response.php
@@ -190,13 +190,15 @@ class Response extends AbstractMessage implements ResponseInterface
         $firstLine = array_shift($lines);
 
         $response = new static();
+        $response->parseStatusLine($firstLine);
 
-        try {
-            $response->parseStatusLine($firstLine);
-        } catch (RuntimeException $e) {
-            array_shift($lines); // skip next empty line
-            $firstLine = array_shift($lines); // and try again
-            $response->parseStatusLine($firstLine);
+        /**
+         * @link https://tools.ietf.org/html/rfc7231#section-6.2.1
+         */
+        if ($response->statusCode === static::STATUS_CODE_100) {
+            $next = array_shift($lines); // take next line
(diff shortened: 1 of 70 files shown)


t1gor:
The build fails with an error irrelent to the change:
> Parse error: syntax error, unexpected ':', expecting ';' or '{' in /home/travis/build/zendframework/zend-http/vendor/doctrine/instantiator/src/Doctrine/Instantiator/Instantiator.php on line 95

src/Response.php
```diff
@@ -180,18 +181,14 @@ public static function fromString($string)
...
-        if (! preg_match($regex, $firstLine, $matches)) {
-            throw new Exception\InvalidArgumentException(
-                'A valid response status line was not found in the provided string'
-            );
+        try {
+            $response->parseStatusLine($firstLine);
+        } catch (RuntimeException $e) {
+            array_shift($lines); // skip next empty line
+            $firstLine = array_shift($lines); // and try again
+            $response->parseStatusLine($firstLine);
```
Xerkus: Make sure next line is empty before skipping.
On a quick glance at the specification I do not see anything preventing sending more headers with 100 status code before proceeding with actual response.

Also your extracted method does too much. Extract status line parsing logic but leave the rest out of it.
I would do something like this:

```php
list($version, $status, $reason) = $this->parseStatusLine($firstLine);
if (static::STATUS_CODE_100 === $status) {
    $next = array_shift($lines);
    if (!empty($next)) {
        // throw exception, probably
    }
    $next = array_shift($lines);
    list($version, $status, $reason) = $this->parseStatusLine($next);
}

$response->version = $version;
$response->setStatusCode($status);
$response->setReasonPhrase($reason);
```

t1gor: Changed to work without exceptions. Rebased and fixed tests.

weierophinney:
I have cherry-picked the relevant commits, ensuring the changes to tests are not propagated. Merged to master for next release.

Issue #123, opened 2017-05-24
Https request validation


diemax:
Hi ZF team! Just a little thing:

How can I check if the `request` scheme is `https`? 
Regardless the `getScheme()` method, does the `Request` class has a method for that?
I can get the scheme:
```php
$request = $this->request->getUri();
return ($request->getScheme() === 'https`);
```
So, where should I add this validation for make it available for all controllers?
Anyways, would be great add this method on the `Request` object.
```php
$this->request->isSecure() 
// or maybe
$this->request->isHttps()
```

Thanks! great project :1st_place_medal:

Ocramius:
What's wrong with `strtolower($uri->getScheme()) === 'https'`? Also, note that the URI does not involve "secure", as the certificate validation may have been disabled on the HTTP client.

diemax:
Cool, nothing wrong with that validation, just I need to know where can I put the function and make it available for all controllers!
Thanks

Ocramius:
@diemax if you are using `zend-mvc`, then a tiny injectable service would be best, but it's also simpler to enforce `https` [via routing and the `Scheme` route](https://github.com/zendframework/zend-router/blob/f75dee88dde6b2cebc123a225be0c2b17015ffdf/src/Http/Scheme.php#L18)

boesing:
Or just create an own controller plugin.
More informations can be found in the `zend-mvc` documentation.
https://docs.zendframework.com/zend-mvc/plugins/

There is no proper documentation, but you can add controller plugins to your configuration by using the `controller_plugins` identifier, which is a configuration namespace for the `ControllerPluginManager`.

```
return [
    'controller_plugins' => [
        'factories' => [
              MyPlugin::class => InvokableFactory::class,
        ],
        'aliases' => [
            'isSecureRequest' => MyPlugin::class,
        ],
    ],
]; 
```

With that, you can execute `$this->isSecureRequest()` in any *Controller*.

If you dont want to have such overhead, you can just create a trait and use it in your controllers:
```
class FooController extends AbstractActionController
{
     use SecureRequestTrait;
    
     public function indexAction() 
     {
          if ($this->isSecureRequest($this->getRequest())) {
               // Whatever
         }
     }
}

trait SecureRequestTrait
{
     protected function isSecureRequest(RequestInterface $request)
     {
          if (!$request instanceof \Zend\Http\Request) {
              return false;
          }
         return strtolower($request->getUri()->getScheme()) === 'https';
     }
}
```

[...]

PR #134, opened 2017-10-24
Added support for x-forwarded-proto header to request object

SteveReiter:


SteveReiter:
Relates to https://github.com/zendframework/zend-http/issues/123

src/HttpsUri.php
```diff
@@ -0,0 +1,23 @@
...
+ * @see       https://github.com/zendframework/zend-http for the canonical source repository
+ * @copyright Copyright (c) 2005-2017 Zend Technologies USA Inc. (http://www.zend.com)
+ * @license   https://github.com/zendframework/zend-http/blob/master/LICENSE.md New BSD License
+ */
+
+namespace Zend\Http;
+
+use Zend\Uri\Http as Http;
+
+class HttpsUri extends Http
```
Xerkus: There is no reason to introduce this class.
`$uri->setScheme('https');` provides exactly the same behavior

src/Request.php
```diff
@@ -228,6 +229,11 @@ public function getUri()
         if ($this->uri === null || is_string($this->uri)) {
             $this->uri = new HttpUri($this->uri);
         }
+
+        $headers = $this->getHeaders()->toArray();
+        if (array_key_exists('X-Forwarded-Proto', $headers) && $headers['X-Forwarded-Proto'] === 'https') {
```
Xerkus: This will end up overriding scheme each time Uri is fetched from Request. Getters must not have side effects

src/HttpsUri.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * Get the scheme part of the URI
+     *
+     * @return string|null
+     */
+    public function getScheme()
+    {
+        return 'https';
```
Xerkus: This breaks Uri contract by ignoring scheme changes
Besides, internal logic will be broken in unpredictable ways where scheme property is used directly

src/Request.php
```diff
@@ -228,6 +229,11 @@ public function getUri()
         if ($this->uri === null || is_string($this->uri)) {
             $this->uri = new HttpUri($this->uri);
         }
+
+        $headers = $this->getHeaders()->toArray();
```
Xerkus: You don't need all the headers. Use `has()` and `get()`

weierophinney:
There are multiple ways of doing this.

First, as suggested in #123, you could write a controller plugin for this. In that case, you can look not only at the scheme of the URI composed by the request, but also at the `X-Forwarded-Proto` header to determine what to return.

Second, you could create a listener that triggers at either low priority in the `bootstrap` event or high priority in the `route` event that resets the URI scheme based on the presence and value of the `X-Forwarded-Proto` header.

Third, we could directly support this via the `Zend\Http\PhpEnvironment\Request` object, and, if that header is detected during initialization, set the URI scheme based on it.

All three of these are preferable to the approach you are presenting here. As @Xerkus notes, the changes you are introducing mean that users would be surprised if the scheme did not update. Additionally, it still requires the user to update their system to inject the alternate URI instance — and if you have to do that, one of the above approaches will be just as successful.

I will consider an approach in this package that implements the third option I outline above, or an approach in the zend-mvc pacakge that implements the second option above.

weierophinney:
Closing due to inactivity. Re-submit if you are able to implement one of the options outlined in my last comment above.

intellent:
Just one addition to this. We use a nginx reverse proxy to terminate SSL and the X-Forwarded-Proto header is not available in our setup. Instead `$_SERVER['HTTP_FRONT_END_HTTPS']` is set to "on".

PR #135, opened 2017-10-25
Added certainty to suggest

carnage:
Adds https://github.com/paragonie/certainty to the composer suggest. 

Will be playing with a PR to add direct support to the adaptors but I consider this as a minimum step to help  users write secure code.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,9 @@
         "zendframework/zend-coding-standard": "~1.0.0",
         "zendframework/zend-config": "^3.1 || ^2.6"
     },
+    "suggest": {
+        "paragonie/certainty": "For automated management of cacert.pem"
+    },
     "autoload": {
         "psr-4": {
             "Zend\\Http\\": "src/"


Xerkus:
It will be 👎 from me.
While I know of the guy and I trust he have security expertise, I cannot put trust into his keys for something as important as certificate authority keys. For the same reason I am against well known framework endorsing such trust for its users.

asgrim:
Interesting, and good point there @Xerkus - I'd be interested to hear from @ezimuel on this one.

From my perspective, it's about trust, and by adding the `suggest`, it's essentially an endorsement. I know Scott is doing a stellar job of pushing security in PHP, and this is a good thing to make things simpler...

carnage:
It is all about trust, but the question becomes who should you trust with this responsibility? Your OS, your web host? Perhaps Zend could maintain a fork signed with their own keys?

Xerkus:
@carnage In the environment I can control it is part of infrastructure automation. Using mozilla as a source with the help of curl provided script to fetch and convert certdata.  In fedora ca-certificates package is good enough for me - same mozilla CA bundle, but updated twice a year.
In the environment I can't control... is that still a thing in 2017?

Result is pretty much the same as with `certainty`, except no middleman to rely on, it is done outside of php and resulting bundle file is not writable by php user.

carnage:
Another question to ask is:

Given that some users do not do what your suggesting due to lack of knowledge or time, which protects them more: them disabling CA verification because it breaks their app or them using certainty to manage that problem for them?

PR #136, opened 2017-10-29
Remove duplicated v2.6.0 from changelog

samizdam:


Xerkus:
Good catch, must be wrong merge conflict resolution.

The other one is wrong tho. Fixed in master

PR #140, opened 2018-01-11
Test if headers return last added header value

thestanislav:
If you add same header multiple times with different values, you will not be able to get last added header value;


weierophinney:
Thanks for the test, @thestanislav! I was able to find a workable solution, and will release shortly.

PR #141, opened 2018-01-25
Fluent Stream::setCleanup()

lionelgaillard:
Because `Stream::setCleanup()` was not returning `$this`, it was breaking the fluent interface.

### DIFF
--- a/src/Response/Stream.php
+++ b/src/Response/Stream.php
@@ -113,6 +113,7 @@ class Stream extends Response
     public function setCleanup($cleanup = true)
     {
         $this->cleanup = $cleanup;
+        return $this;
     }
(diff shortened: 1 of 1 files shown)


webimpress:
@lionelgaillard As I remember we are dropping fluent interfaces, please see:
https://discourse.zendframework.com/t/how-should-we-provide-return-type-declarations-for-fluent-interfaces/30/13
and @Ocramius blog post: https://ocramius.github.io/blog/fluent-interfaces-are-evil/

Please note that also `setContentLength` method (for example) does not return `$this`...

lionelgaillard-inexto:
Ho ok...

PR #143, opened 2018-04-16
Updated Travis CI configuration - PHP 7.2

webimpress:
- removed allow_failures on PHP 7.2
- removed composer self-update
- updated legacy deps
- simplified disabling xdebug
- moved cs-check and coverage test to run on PHP 7.1 with locked deps


weierophinney:
Thanks, @webimpress.

PR #148, opened 2018-05-06
Client\Adapter\Proxy, Socket, and Test optimization on write() method to avoid repetitive uri->getQuery() when possible

samsonasik:

### DIFF
--- a/src/Client/Adapter/Proxy.php
+++ b/src/Client/Adapter/Proxy.php
@@ -156,8 +156,8 @@ class Proxy extends Socket
         // Build request headers
         if ($this->negotiated) {
             $path = $uri->getPath();
-            if ($uri->getQuery()) {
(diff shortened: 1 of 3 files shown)


src/Client/Adapter/Proxy.php
```diff
@@ -156,8 +156,8 @@ public function write($method, $uri, $httpVer = '1.1', $headers = [], $body = ''
         // Build request headers
         if ($this->negotiated) {
             $path = $uri->getPath();
-            if ($uri->getQuery()) {
-                $path .= '?' . $uri->getQuery();
+            if ($query = $uri->getQuery()) {
```
weierophinney: I really don't like doing assignment in conditionals, as it's an easy vector for mistakes. When I merge, I'll assign prior to the conditional.

weierophinney:
Thanks, @samsonasik

PR #149, opened 2018-05-06
Fixes #36 Client::setUri() should keep relative uri status

### LINKED ISSUE #36: setUri() invalidates relative URI when setting default ports
Hi,

when setting a relative URI `/example` the `Zend\Http\Client::setUri()` method adds a default port and so invalidates the URI resulting in isValidRelative() returning FALSE instead of TRUE.

``` php
<?php
require './vendor/autoload.php';
$relativeUri = '/example';
$client = new Zend\Http\Client($relativeUri);
$uri = new Zend\Uri\Http($relativeUri);
var_dump($client->getUri()->isValidRelative(), $uri->isValidRelative());
```

_Expected output_

```
boolean(true)
boolean(true)
```

_Actual output_

```
boolean(false)
boolean(true)
```

I don't know how to build a patch, but line 323 of `Client.php` should be
`if (! $this->getUri()->getPort() && $this->getUri()->isAbsolute()) {`
instead of
`if (! $this->getUri()->getPort()) {`

Cheers

samsonasik:
Provide a narrative description of what you are trying to accomplish:

- [x] Are you fixing a bug?
          Supply relative uri to `Client::setUri()`
          It convert to absolute uri
          It should keep as relative uri 
  - [x] Add a `CHANGELOG.md` entry for the fix.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -327,13 +327,13 @@ class Client implements Stdlib\DispatchableInterface
             $uri = $this->getUri();
 
             // Set auth if username and password has been specified in the uri
-            if ($user = $uri->getUser() && $password = $uri->getPassword()) {
+            if (($user = $uri->getUser()) && ($password = $uri->getPassword())) {
                 $this->setAuth($user, $password);
             }
 
             // We have no ports, set the defaults
             if (! $uri->getPort() && $uri->isAbsolute()) {
-                $uri->setPort(($uri->getScheme() == 'https' ? 443 : 80));
+                $uri->setPort($uri->getScheme() === 'https' ? 443 : 80);
             }
         }
         return $this;
@@ -904,7 +904,7 @@ class Client implements Stdlib\DispatchableInterface
             }
             // If we have no ports, set the defaults
             if (! $uri->getPort() && $uri->isAbsolute()) {
-                $uri->setPort($uri->getScheme() == 'https' ? 443 : 80);
+                $uri->setPort($uri->getScheme() === 'https' ? 443 : 80);
             }
 
             // method
--- a/src/Client.php
+++ b/src/Client.php
@@ -324,14 +324,16 @@ class Client implements Stdlib\DispatchableInterface
                 $this->clearAuth();
             }
 
+            $uri = $this->getUri();
+
             // Set auth if username and password has been specified in the uri
-            if ($this->getUri()->getUser() && $this->getUri()->getPassword()) {
-                $this->setAuth($this->getUri()->getUser(), $this->getUri()->getPassword());
+            if ($user = $uri->getUser() && $password = $uri->getPassword()) {
+                $this->setAuth($user, $password);
             }
 
             // We have no ports, set the defaults
-            if (! $this->getUri()->getPort() && $this->getUri()->isAbsolute()) {
-                $this->getUri()->setPort(($this->getUri()->getScheme() == 'https' ? 443 : 80));
+            if (! $uri->getPort() && $uri->isAbsolute()) {
+                $uri->setPort(($uri->getScheme() == 'https' ? 443 : 80));
             }
         }
         return $this;
--- a/src/Client.php
+++ b/src/Client.php
@@ -330,7 +330,7 @@ class Client implements Stdlib\DispatchableInterface
             }
 
             // We have no ports, set the defaults
-            if (! $this->getUri()->getPort()) {
+            if (! $this->getUri()->getPort() && $this->getUri()->isAbsolute()) {
                 $this->getUri()->setPort(($this->getUri()->getScheme() == 'https' ? 443 : 80));
             }
         }
--- a/src/Client.php
+++ b/src/Client.php
@@ -903,7 +903,7 @@ class Client implements Stdlib\DispatchableInterface
                 }
             }
             // If we have no ports, set the defaults
-            if (! $uri->getPort()) {
+            if (! $uri->getPort() && $uri->isAbsolute()) {
                 $uri->setPort($uri->getScheme() == 'https' ? 443 : 80);
             }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -325,9 +325,11 @@ class Client implements Stdlib\DispatchableInterface
             }
 
             $uri = $this->getUri();
+            $user = $uri->getUser();
+            $password = $uri->getPassword();
 
             // Set auth if username and password has been specified in the uri
-            if (($user = $uri->getUser()) && ($password = $uri->getPassword())) {
+            if ($user && $password) {
                 $this->setAuth($user, $password);
             }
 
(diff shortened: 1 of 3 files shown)


samsonasik:
After some check, I think relative url should not be allowed instead in `Zend\Http\Client`, so, instead of keeping it, we can do these options:

1. throw Exception to warn user to use absolute URI, OR
2. set URI to use current host detected as `$lastHost = $this->getRequest()->getUri()->getHost();`

src/Client.php
```diff
@@ -324,14 +324,16 @@ public function setUri($uri)
                 $this->clearAuth();
             }
 
+            $uri = $this->getUri();
+
             // Set auth if username and password has been specified in the uri
-            if ($this->getUri()->getUser() && $this->getUri()->getPassword()) {
-                $this->setAuth($this->getUri()->getUser(), $this->getUri()->getPassword());
+            if ($user = $uri->getUser() && $password = $uri->getPassword()) {
```
webimpress: We need brackets around expressions here, see following example - https://3v4l.org/oV5F5:
```php
<?php

function a() {
    return 1;
}

function b() {
    return 2;
}

if ($a = a() && $b = b()) {
    var_dump($a, $b);
}
```

results:
```
bool(true)
int(2)
```

and because of that I think there is missing proper test for that change.

src/Client.php
```diff
@@ -324,14 +324,16 @@ public function setUri($uri)
...
+                $this->setAuth($user, $password);
             }
 
             // We have no ports, set the defaults
-            if (! $this->getUri()->getPort()) {
-                $this->getUri()->setPort(($this->getUri()->getScheme() == 'https' ? 443 : 80));
+            if (! $uri->getPort() && $uri->isAbsolute()) {
+                $uri->setPort(($uri->getScheme() == 'https' ? 443 : 80));
```
webimpress: I know it was like that before, but we can do two changes here:
- remove redundant bracket
- change `==` to `===`

test/ClientTest.php
```diff
@@ -524,4 +524,28 @@ public function testFormUrlEncodeSeparator()
...
+    {
+        $client = new Client($uri);
+        $this->assertSame($isValidRelativeURI, $client->getUri()->isValidRelative());
+    }
+
+    public function testSendRequestWithRelativeURI()
+    {
+        $client = new Client('/example');
+        $client->setAdapter(Test::class);
+        $client->send();
```
webimpress: There is no any assertion in that test. It is really unclear what we are testing here.

samsonasik: done ;)

src/Client.php
samsonasik: done ;)

src/Client.php
samsonasik: done ;)

samsonasik:
@webimpress I've added the tests for check `isAbsolute()`

src/Client.php
```diff
@@ -324,14 +324,16 @@ public function setUri($uri)
                 $this->clearAuth();
             }
 
+            $uri = $this->getUri();
+
             // Set auth if username and password has been specified in the uri
-            if ($this->getUri()->getUser() && $this->getUri()->getPassword()) {
-                $this->setAuth($this->getUri()->getUser(), $this->getUri()->getPassword());
+            if (($user = $uri->getUser()) && ($password = $uri->getPassword())) {
```
weierophinney: Please don't do assignments in conditionals.

test/ClientTest.php
```diff
@@ -524,4 +524,59 @@ public function testFormUrlEncodeSeparator()
...
+        return [
+            ['/example', true],
+            ['http://localhost/example', false]
+        ];
+    }
+
+    /**
+     * @dataProvider uriDataProvider
+     */
+    public function testValidRelativeURI($uri, $isValidRelativeURI)
```
weierophinney: What _behavior_ is this testing, exactly? This should likely be "testUriCorrectlyDeterminesWhetherOrNotItIsAValidRelativeUri".

test/ClientTest.php
```diff
@@ -524,4 +524,59 @@ public function testFormUrlEncodeSeparator()
         $rawRequest = $client->getLastRawRequest();
         $this->assertContains('foo=bar&baz=foo', $rawRequest);
     }
+
+    public function uriDataProvider()
+    {
+        return [
+            ['/example', true],
+            ['http://localhost/example', false]
+        ];
```
weierophinney: Please add _names_ for each case in the provider. This makes it simpler to determine which case failed when a failure occurs. As an example:

```php
'valid-relative' => ['/example', true],
'invalid-absolute' => ['http://localhost/example', false],
```

test/ClientTest.php
```diff
@@ -524,4 +524,59 @@ public function testFormUrlEncodeSeparator()
...
+        $client->send();
+        $this->assertSame($isValidRelativeURI, $client->getUri()->isValidRelative());
+    }
+
+    public function portChangeDataProvider()
+    {
+        return [
+            ['https://localhost/example', 443],
+            ['http://localhost/example', 80]
+        ];
```
weierophinney: Same comment here about adding names for each provider case.

test/ClientTest.php
```diff
@@ -524,4 +524,59 @@ public function testFormUrlEncodeSeparator()
...
+    }
+
+    /**
+     * @dataProvider portChangeDataProvider
+     */
+    public function testAbsoluteSetPortWhenNoPort($absoluteURI, $port)
```
weierophinney: This needs a name that better describes the behavior being tested: "testUriPortIsSetToAppropriateDefaultValueWhenAnAbsoluteUriOmittingThePortIsProvided".

test/ClientTest.php
```diff
@@ -524,4 +524,59 @@ public function testFormUrlEncodeSeparator()
...
+
+        $client->setUri($absoluteURI);
+        $this->assertSame($port, $client->getUri()->getPort());
+
+        $client->setAdapter(Test::class);
+        $client->send();
+        $this->assertSame($port, $client->getUri()->getPort());
+    }
+
+    public function testRelativeURIDoesnotSetPort()
```
weierophinney: Need a better method name here; I'll leave this as an exercise for you!

samsonasik:
@weierophinney done ;). All incorporated.

weierophinney:
@samsonasik https://travis-ci.org/zendframework/zend-http/jobs/415602200#L792

samsonasik:
@weierophinney should be fixed now

samsonasik:
travis green

samsonasik:
@weierophinney mergeable ?

weierophinney:
Thanks, @sunspikes!

samsonasik:
it make keep relative uri status, for host exists check, I already commented before at https://github.com/zendframework/zend-http/pull/149#issuecomment-386852954 but that will not related to this PR, keeping exisitng URI host for relative uri seems need another PR

PR #151, opened 2018-05-30
Passing an invalid uri to the referrer header should no longer throw

roelvanduijnhoven:
Due to the change in https://github.com/zendframework/zend-http/pull/147 we removed the try/catch control structures from our code. However: exceptions are still generated. I think that this is not intended.

The only valid way to check for a referrer is now the following. Which is kind of cumbersome:
```php
function isRefererFromGitHub($headers) {
    try {
        $referer = $headers->get('Referer');
        if (!$referer instanceof \Zend\Http\Header\GenericHeader) {
            return false;
        }
        
        return $referer->getUri()->getHost() === 'github.com';
    } catch (\Zend\Http\Header\Exception\InvalidArgumentException $e) {
        return false;
    }
}
```

This PR fixes this by wrapping the exception generated. 

----

I am however a bit unsure if this is correct. The release notes state:

> Now, Header::has() will return false for such headers

However; in my case the `has` function _does_ yield true. Therefore I omitted that assertion from my test code.

Any thoughts @webimpress @weierophinney?

### DIFF
--- a/src/Header/AbstractLocation.php
+++ b/src/Header/AbstractLocation.php
@@ -76,6 +76,12 @@ abstract class AbstractLocation implements HeaderInterface
                     $e->getCode(),
                     $e
                 );
+            } catch (UriException\InvalidArgumentException $e) {
+                throw new Exception\InvalidArgumentException(
+                    sprintf('Invalid URI passed as string (%s)', (string) $uri),
+                    $e->getCode(),
+                    $e
+                );
             }
         } elseif (! ($uri instanceof UriInterface)) {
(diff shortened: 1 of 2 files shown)


weierophinney:
This is a case where `has()` SHOULD return `true`, but `get()` MUST raise an exception. As such.

We have a similar situation in zend-servicemanager: an entry can be both registered, allowing `has()` to return `true`, but, due to missing dependencies or other problems, `get()` will raise an exception. This is normal; `has() ` is simply indicating that the container knows about the entry. As such, your changes make sense.

weierophinney:
Thanks, @roelvanduijnhoven!

PR #152, opened 2018-06-22
using "Cookie|false" instead of "Cookie|bool" for docblock Request::getCookie()

samsonasik:
- [x] Is this related to documentation?

As when no "Cookie" header, it will return false.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -314,7 +314,7 @@ class Request extends AbstractMessage implements RequestInterface
      * Return the Cookie header, this is the same as calling $request->getHeaders()->get('Cookie');
      *
      * @convenience $request->getHeaders()->get('Cookie');
-     * @return Header\Cookie|bool
+     * @return Header\Cookie|false
      */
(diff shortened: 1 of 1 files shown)


src/Request.php
```diff
@@ -314,7 +314,7 @@ public function getPost($name = null, $default = null)
      * Return the Cookie header, this is the same as calling $request->getHeaders()->get('Cookie');
      *
      * @convenience $request->getHeaders()->get('Cookie');
-     * @return Header\Cookie|bool
+     * @return Header\Cookie|false
```
Ocramius: `false` is not a type: it's a value.

This change is invalid, sorry.

samsonasik: @Ocramius Type can be explicit boolean value based on it http://docs.phpdoc.org/guides/types.html

Ocramius: That's still not a type - even if phpdocumentor supports it, that is a mistake

PR #153, opened 2018-07-17
update reason phrase for 425 response status code to "Too Early"

samsonasik:
refs https://github.com/zendframework/zend-diactoros/pull/313


weierophinney:
Thanks, @samsonasik

Issue #127, opened 2017-08-14
Set-Cookie: need ability to set value without URL encoding


ademarre:
Zend\Http\Header\SetCookie should allow setting the value without URL encoding it, similar to PHP's setrawcookie() function. This can be necessary for compatibility with other libraries, or when you know that all characters are safe and no escaping is necessary.

For example, a cookie whose value might contain a colon (`:`) is perfectly valid:

    Set-Cookie:foo=a:b; Path=/
But currently it can only be set like this:

    Set-Cookie:foo=a%3Ab; Path=/

API consistency considerations:
- Zend\Http\Header\Cookie has a `setEncodeValue()` method that can be used to toggle whether cookie values are URL encoded.
- SetCookie has `setQuoteFieldValue()` and `hasQuoteFieldValue()` methods. This quoting is applied to the same header component as URL encoding. In my opinion, these names are misleading, since the quoting applies only to the cookie value itself, not the entire "field value", as the term is typically used. This is only a minor concern though, as anyone familiar enough with the spec to want quotes probably knows what it does.
- Adding `setEncodeFieldValue()` / `hasEncodeFieldValue()` would be consistent with the precedent set by `setQuoteFieldValue()`, but slightly misleading.
- `setEncodeValue()` / `getEncodeValue()` are more accurate names and are consistent with the Cookie class.
- `setEncodeValue()` / `hasEncodeValue()` is consistent with the "set" / "has" terminology used for SetCookie's quote methods, but doesn't make the mistake of implying the entire field value is encoded.

PR #154, opened 2018-10-13
#127 Set-Cookie: added ability to set value without URL encoding

thomasvargiu:
#127 
- Added `Added Zend\Http\Header\SetCookie::setEncodeValue()` and `Added Zend\Http\Header\SetCookie::setEncodeValue()` to allow set value without url encoding


weierophinney:
Thanks, @thomasvargiu! Merged to develop for release with 2.9.0.

Issue #76, opened 2016-06-28
Response class blindly attempts to inflate empty body


GeeH:
This issue has been moved from the `zendframework` repository as part of the bug migration program as outlined here - http://framework.zend.com/blog/2016-04-11-issue-closures.html

---

Original Issue: https://api.github.com/repos/zendframework/zendframework/issues/7220
User: @miguelsimoes
Created On: 2015-02-14T16:44:51Z
Updated At: 2015-11-06T20:31:08Z
Body
Looking at the code:

```
    ErrorHandler::start();
    $return = gzinflate(substr($body, 10));
    $test = ErrorHandler::stop();
    if ($test) {
        throw new Exception\RuntimeException(
            'Error occurred during gzip inflation',
            0,
            $test
        );
    }
```

That exists in the decodeGzip method under the \Zend\HttpResponse FQCN, this breaks when you get a Content-Encoding: gzip header with empty body. The Response class should verify the Content-Length associated with the response before blindly attempt to call the decodeGzip method.

---

YuraKril:
Are there any updates?

Ocramius:
@YuraKril if you experience this as well, please open a PR with a failing test case.

PR #156, opened 2018-10-13
#76 Check content-length before to decode response body

thomasvargiu:
With this PR we check the content-length header, if it's 0 we didn't try to decode content body.


weierophinney:
Thanks, @thomasvargiu!

weierophinney:
Merged to develop for release with 2.9.0.

PR #157, opened 2018-10-16
phpstan fixes

thomasvargiu:
- some fixes on PHPDoc 
- added checks on types when needed

### Changed

- Throw an exception when unable to uncompress response content
- Relative URIs are not allowed anymore in `Zend\Http\Client`.
  Anyway, usage of relative URIs has no sense and no adapter can handle it.

### Fixed

- Fixed `Zend\Http\Cookies::getAllCookies(Zend\Http\Cookies::COOKIE_STRING_ARRAY)`.
- Fixed `Zend\Http\Cookies::getAllCookies(Zend\Http\Cookies::COOKIE_STRING_CONCAT)`.
- Fixed use of HTTP URIs with empty path in Request

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   global:
     - COMPOSER_ARGS="--no-interaction"
     - COVERAGE_DEPS="php-coveralls/php-coveralls"
+    - PHPSTAN_DEPS="phpstan/phpstan:^0.12"
     - TESTS_ZEND_HTTP_CLIENT_ONLINE=true
     - TESTS_ZEND_HTTP_CLIENT_BASEURI=http://127.0.0.1
     - TESTS_ZEND_HTTP_CLIENT_HTTP_PROXY=127.0.0.1:8081
@@ -43,6 +44,7 @@ matrix:
         - DEPS=locked
         - CS_CHECK=true
         - TEST_COVERAGE=true
+        - PHPSTAN_TEST=true
     - php: 7.1
       env:
         - DEPS=latest
@@ -74,6 +76,7 @@ install:
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update $COMPOSER_ARGS ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable $COMPOSER_ARGS ; fi
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS $COVERAGE_DEPS ; fi
+  - if [[ $PHPSTAN_TEST == 'true' ]]; then travis_retry composer require --dev $COMPOSER_ARGS $PHPSTAN_DEPS ; fi
   - stty cols 120 && composer show
 
 before_script:
@@ -105,6 +108,7 @@ before_script:
 script:
   - if [[ $TEST_COVERAGE == 'true' ]]; then composer test-coverage ; else composer test ; fi
   - if [[ $CS_CHECK == 'true' ]]; then composer cs-check ; fi
+  - if [[ $PHPSTAN_TEST == 'true' ]]; then ./vendor/bin/phpstan analyse --no-progress . ; fi
 
 after_script:
   - if [[ $TEST_COVERAGE == 'true' ]]; then travis_retry php vendor/bin/php-coveralls -v ; fi
--- a/src/Client.php
+++ b/src/Client.php
@@ -331,6 +331,10 @@ class Client implements Stdlib\DispatchableInterface
             // cleared for peer subdomains due to technical limits
             $nextHost = $this->getRequest()->getUri()->getHost();
 
+            if (! $nextHost) {
+                throw new InvalidArgumentException('Relative URIs are not allowed');
+            }
+
             if (! preg_match('/' . preg_quote($lastHost, '/') . '$/i', $nextHost)) {
                 $this->clearAuth();
             }
@@ -809,7 +813,7 @@ class Client implements Stdlib\DispatchableInterface
      * @param string $user
      * @param string $password
      * @param string $type
-     * @param array $digest
+     * @param string[] $digest
      * @param null|string $entityBody
      * @throws Exception\InvalidArgumentException
      * @return string|bool
@@ -838,7 +842,7 @@ class Client implements Stdlib\DispatchableInterface
                     throw new Exception\InvalidArgumentException('The digest cannot be empty');
                 }
                 foreach ($digest as $key => $value) {
-                    if (! defined('self::DIGEST_' . strtoupper($key))) {
+                    if (! defined('self::DIGEST_' . strtoupper((string) $key))) {
                         throw new Exception\InvalidArgumentException(sprintf(
                             'Invalid or not supported digest authentication parameter: \'%s\'',
                             $key
@@ -917,7 +921,7 @@ class Client implements Stdlib\DispatchableInterface
 
                 if (! empty($queryArray)) {
                     $newUri = $uri->toString();
-                    $queryString = http_build_query($queryArray, null, $this->getArgSeparator());
+                    $queryString = http_build_query($queryArray, '', $this->getArgSeparator());
 
                     if ($this->config['rfc3986strict']) {
                         $queryString = str_replace('+', '%20', $queryString);
@@ -1324,7 +1328,8 @@ class Client implements Stdlib\DispatchableInterface
                 $this->setEncType(self::ENC_FORMDATA);
             }
         } else {
-            $this->setEncType($this->getHeader('Content-Type') ?: null);
+            $contentType = $this->getHeader('Content-Type');
+            $this->setEncType(\is_string($contentType) && ! empty($contentType) ? $contentType: null);
         }
 
         // If we have POST parameters or files, encode and add them to the body
@@ -1353,7 +1358,7 @@ class Client implements Stdlib\DispatchableInterface
                 $body .= '--' . $boundary . '--' . "\r\n";
             } elseif (stripos($this->getEncType(), self::ENC_URLENCODED) === 0) {
                 // Encode body as application/x-www-form-urlencoded
-                $body = http_build_query($this->getRequest()->getPost()->toArray(), null, '&');
+                $body = http_build_query($this->getRequest()->getPost()->toArray(), '', '&');
             } else {
                 throw new Client\Exception\RuntimeException(sprintf(
                     'Cannot handle content type \'%s\' automatically',
--- a/src/Client.php
+++ b/src/Client.php
@@ -1329,7 +1329,7 @@ class Client implements Stdlib\DispatchableInterface
             }
         } else {
             $contentType = $this->getHeader('Content-Type');
-            $this->setEncType(\is_string($contentType) && ! empty($contentType) ? $contentType: null);
+            $this->setEncType(\is_string($contentType) ? $contentType : '');
         }
 
         // If we have POST parameters or files, encode and add them to the body
--- a/src/Client.php
+++ b/src/Client.php
@@ -1028,7 +1028,7 @@ class Client implements Stdlib\DispatchableInterface
 
             // Get the cookies from response (if any)
             $setCookies = $response->getCookie();
-            if (! \is_bool($setCookies) && ! empty($setCookies)) {
+            if (! is_bool($setCookies) && ! empty($setCookies)) {
                 $this->addCookie($setCookies);
             }
 
@@ -1329,7 +1329,7 @@ class Client implements Stdlib\DispatchableInterface
             }
         } else {
             $contentType = $this->getHeader('Content-Type');
-            $this->setEncType(\is_string($contentType) ? $contentType : '');
+            $this->setEncType(is_string($contentType) ? $contentType : '');
         }
 
         // If we have POST parameters or files, encode and add them to the body
--- a/src/Client.php
+++ b/src/Client.php
@@ -11,6 +11,11 @@ use ArrayIterator;
 use Traversable;
 use Zend\Http\Client\Adapter\Curl;
 use Zend\Http\Client\Adapter\Socket;
+use Zend\Http\Client\Adapter\StreamInterface;
+use Zend\Http\Exception\InvalidArgumentException;
+use Zend\Http\Exception\RuntimeException;
+use Zend\Http\Exception\UnexpectedValueException;
+use Zend\Http\Header\HeaderInterface;
 use Zend\Http\Header\SetCookie;
 use Zend\Stdlib;
 use Zend\Stdlib\ArrayUtils;
@@ -45,17 +50,17 @@ class Client implements Stdlib\DispatchableInterface
     const DIGEST_CNONCE = 'cnonce';
 
     /**
-     * @var Response
+     * @var null|Response
      */
     protected $response;
 
     /**
-     * @var Request
+     * @var null|Request
      */
     protected $request;
 
     /**
-     * @var Client\Adapter\AdapterInterface
+     * @var null|Client\Adapter\AdapterInterface
      */
     protected $adapter;
 
@@ -65,14 +70,14 @@ class Client implements Stdlib\DispatchableInterface
     protected $auth = [];
 
     /**
-     * @var string
+     * @var null|string|resource
      */
     protected $streamName;
 
     /**
      * @var resource|null
      */
-    protected $streamHandle = null;
+    protected $streamHandle;
 
     /**
      * @var array of Header\SetCookie
@@ -85,12 +90,12 @@ class Client implements Stdlib\DispatchableInterface
     protected $encType = '';
 
     /**
-     * @var Request
+     * @var null|string
      */
     protected $lastRawRequest;
 
     /**
-     * @var Response
+     * @var null|string
      */
     protected $lastRawResponse;
 
@@ -128,15 +133,15 @@ class Client implements Stdlib\DispatchableInterface
      * This variable is populated the first time _detectFileMimeType is called
      * and is then reused on every call to this method
      *
-     * @var resource
+     * @var null|resource
      */
     protected static $fileInfoDb;
 
     /**
      * Constructor
      *
-     * @param string $uri
-     * @param array|Traversable $options
+     * @param null|string $uri
+     * @param null|array|Traversable $options
      */
     public function __construct($uri = null, $options = null)
     {
@@ -170,9 +175,7 @@ class Client implements Stdlib\DispatchableInterface
         }
 
         // Pass configuration options to the adapter if it exists
-        if ($this->adapter instanceof Client\Adapter\AdapterInterface) {
-            $this->adapter->setOptions($options);
-        }
+        $this->getAdapter()->setOptions($options);
 
         return $this;
     }
@@ -203,9 +206,12 @@ class Client implements Stdlib\DispatchableInterface
         }
 
         $this->adapter = $adapter;
+
         $config = $this->config;
         unset($config['adapter']);
-        $this->adapter->setOptions($config);
+
+        $adapter->setOptions($config);
+
         return $this;
     }
 
@@ -220,7 +226,10 @@ class Client implements Stdlib\DispatchableInterface
             $this->setAdapter($this->config['adapter']);
         }
 
-        return $this->adapter;
+        /** @var Client\Adapter\AdapterInterface $adapter */
+        $adapter = $this->adapter;
+
+        return $adapter;
     }
 
     /**
@@ -277,7 +286,7 @@ class Client implements Stdlib\DispatchableInterface
     /**
      * Get the last request (as a string)
      *
-     * @return string
+     * @return null|string
      */
     public function getLastRawRequest()
     {
@@ -287,7 +296,7 @@ class Client implements Stdlib\DispatchableInterface
     /**
      * Get the last response (as a string)
      *
-     * @return string
+     * @return null|string
      */
     public function getLastRawResponse()
     {
@@ -314,13 +323,14 @@ class Client implements Stdlib\DispatchableInterface
     {
         if (! empty($uri)) {
             // remember host of last request
-            $lastHost = $this->getRequest()->getUri()->getHost();
+            $lastHost = $this->getRequest()->getUri()->getHost() ?: '';
             $this->getRequest()->setUri($uri);
 
             // if host changed, the HTTP authentication should be cleared for security
             // reasons, see #4215 for a discussion - currently authentication is also
             // cleared for peer subdomains due to technical limits
             $nextHost = $this->getRequest()->getUri()->getHost();
+
             if (! preg_match('/' . preg_quote($lastHost, '/') . '$/i', $nextHost)) {
                 $this->clearAuth();
             }
@@ -412,7 +422,7 @@ class Client implements Stdlib\DispatchableInterface
     {
         $argSeparator = $this->config['argseparator'];
         if (empty($argSeparator)) {
-            $argSeparator = ini_get('arg_separator.output');
+            $argSeparator = ini_get('arg_separator.output') ?: '&';
             $this->setArgSeparator($argSeparator);
         }
         return $argSeparator;
@@ -421,14 +431,14 @@ class Client implements Stdlib\DispatchableInterface
     /**
      * Set the encoding type and the boundary (if any)
      *
-     * @param string $encType
-     * @param string $boundary
+     * @param null|string $encType
+     * @param null|string $boundary
      * @return $this
      */
     public function setEncType($encType, $boundary = null)
     {
         if (null === $encType || empty($encType)) {
-            $this->encType = null;
+            $this->encType = '';
             return $this;
         }
 
@@ -490,10 +500,9 @@ class Client implements Stdlib\DispatchableInterface
      * Reset all the HTTP parameters (request, response, etc)
      *
      * @param  bool   $clearCookies  Also clear all valid cookies? (defaults to false)
-     * @param  bool   $clearAuth     Also clear http authentication? (defaults to true)
      * @return $this
      */
-    public function resetParameters($clearCookies = false /*, $clearAuth = true */)
+    public function resetParameters($clearCookies = false)
     {
         $clearAuth = true;
         if (func_num_args() > 1) {
@@ -503,7 +512,7 @@ class Client implements Stdlib\DispatchableInterface
         $uri = $this->getUri();
 
         $this->streamName      = null;
-        $this->encType         = null;
+        $this->encType         = '';
         $this->request         = null;
         $this->response        = null;
         $this->lastRawRequest  = null;
@@ -535,12 +544,12 @@ class Client implements Stdlib\DispatchableInterface
     /**
      * Get the cookie Id (name+domain+path)
      *
-     * @param  Header\SetCookie|Header\Cookie $cookie
+     * @param  Header\SetCookie $cookie
      * @return string|bool
      */
     protected function getCookieId($cookie)
     {
-        if (($cookie instanceof Header\SetCookie) || ($cookie instanceof Header\Cookie)) {
+        if ($cookie instanceof Header\SetCookie) {
             return $cookie->getName() . $cookie->getDomain() . $cookie->getPath();
         }
         return false;
@@ -550,14 +559,14 @@ class Client implements Stdlib\DispatchableInterface
      * Add a cookie
      *
      * @param array|ArrayIterator|Header\SetCookie|string $cookie
-     * @param string  $value
-     * @param string  $expire
-     * @param string  $path
-     * @param string  $domain
-     * @param  bool $secure
-     * @param  bool $httponly
-     * @param string  $maxAge
-     * @param string  $version
+     * @param null|string  $value
+     * @param null|string  $expire
+     * @param null|string  $path
+     * @param null|string  $domain
+     * @param bool $secure
+     * @param bool $httponly
+     * @param null|int  $maxAge
+     * @param null|int  $version
      * @throws Exception\InvalidArgumentException
      * @return $this
      */
@@ -681,12 +690,16 @@ class Client implements Stdlib\DispatchableInterface
     {
         $headers = $this->getRequest()->getHeaders();
 
-        if ($headers instanceof Headers) {
-            if ($headers->get($name)) {
-                return $headers->get($name)->getFieldValue();
-            }
+        if (! $headers instanceof Headers) {
+            return false;
         }
-        return false;
+
+        $header = $headers->get($name);
+        if (! $header instanceof HeaderInterface) {
+            return false;
+        }
+
+        return $header->getFieldValue();
     }
 
     /**
@@ -703,7 +716,7 @@ class Client implements Stdlib\DispatchableInterface
 
     /**
      * Get status of streaming for received data
-     * @return bool|string
+     * @return bool|resource|string
      */
     public function getStream()
     {
@@ -729,7 +742,11 @@ class Client implements Stdlib\DispatchableInterface
             $this->streamName = tempnam(
                 isset($this->config['streamtmpdir']) ? $this->config['streamtmpdir'] : sys_get_temp_dir(),
                 Client::class
-            );
+            ) ?: null;
+
+            if (null === $this->streamName) {
+                throw new RuntimeException('Unable to create temporary name for stream');
+            }
         }
 
         ErrorHandler::start();
@@ -838,6 +855,8 @@ class Client implements Stdlib\DispatchableInterface
                         );
                     }
                     $ha2 = md5($this->getMethod() . ':' . $this->getUri()->getPath() . ':' . md5($entityBody));
+                } else {
+                    throw new InvalidArgumentException('Invalid DIGEST auth data');
                 }
                 if (empty($digest['qop'])) {
                     $response = md5($ha1 . ':' . $digest['nonce'] . ':' . $ha2);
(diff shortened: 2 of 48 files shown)


src/Cookies.php
```diff
@@ -197,7 +199,7 @@ public function getMatchingCookies(
      * @param string $cookieName The cookie's name
      * @param int $retAs Whether to return cookies as objects of \Zend\Http\Header\SetCookie or as strings
      * @throws Exception\InvalidArgumentException if invalid URI specified or invalid $retAs value
-     * @return SetCookie|string
+     * @return SetCookie|string|false
```
Ocramius: `false` is not a type: `bool` is to be used here

src/Cookies.php
```diff
@@ -316,7 +320,9 @@ protected function _matchPath($domains, $path)
         $ret = [];
 
         foreach ($domains as $dom => $pathsArray) {
-            foreach (array_keys($pathsArray) as $cpath) {
+            /** @var string[] $keys */
+            $keys = array_keys($pathsArray);
```
Ocramius: Should be fixable if `$domains` is mapped as `array<string, array<string, mixed>>` (in the parameters)

src/Header/AbstractAccept.php
```diff
@@ -240,9 +241,9 @@ function ($v) {
      *
      * @param  string $type
      * @param  int|float $priority
-     * @param  array (optional) $params
+     * @param  array $params (optional)
      * @throws Exception\InvalidArgumentException
-     * @return Accept
+     * @return $this
```
Ocramius: `$this` is not a type: either `self` or `static`

src/Header/AbstractAccept.php
```diff
@@ -330,9 +331,9 @@ public function match($matchAgainst)
     /**
      * Return a match where all parameters in argument #1 match those in argument #2
      *
-     * @param array $match1
-     * @param array $match2
-     * @return bool|array
+     * @param stdClass $match1
```
Ocramius: Rater than `stdClass`, this should be `object`

src/Header/AbstractAccept.php
```diff
@@ -330,9 +331,9 @@ public function match($matchAgainst)
...
+     * @param stdClass $match2
```
Ocramius: Rater than `stdClass`, this should be `object`

src/Header/AbstractAccept.php
```diff
@@ -330,9 +331,9 @@ public function match($matchAgainst)
...
+     * @return false|stdClass
```
Ocramius: Rater than `stdClass`, this should be `object`

Ocramius: `false` is not a type: `bool` is to be used

src/Header/Age.php
```diff
@@ -73,7 +73,7 @@ public function getFieldValue()
      * Set number of seconds
      *
      * @param int $delta
-     * @return RetryAfter
+     * @return Age
```
Ocramius: `self` is generally better, as it survives renames. Not really important though.

src/Header/ContentType.php
```diff
@@ -238,7 +239,7 @@ protected function assembleValue()
     {
         $mediaType = $this->getMediaType();
         if (empty($this->parameters)) {
-            return $mediaType;
+            return $mediaType ?: '';
```
Ocramius: Since the type is `null|string`, should this be `??`?

src/Header/ContentType.php
```diff
@@ -356,17 +360,17 @@ protected function validateSubtype($right, $left)
      *
      * Validate that the right side format matches what the left side defines.
      *
-     * @param  string $right
-     * @param  string $left
+     * @param  stdClass $right
```
Ocramius: `object`

src/Header/ContentType.php
```diff
@@ -356,17 +360,17 @@ protected function validateSubtype($right, $left)
...
+     * @param  stdClass $left
```
Ocramius: `object`

src/Header/SetCookie.php
```diff
@@ -81,14 +81,19 @@ class SetCookie implements MultipleHeaderInterface
...
     /**
-     * @var bool|null
+     * @var null|bool
      */
     protected $httponly;
 
+    /**
+     * @var string
+     */
+    public $type = 'Cookie';
```
Ocramius: This should probably not be added: it will change how the object behaves on a public API (serialization like `json_encode()`, mostly). What's the reason for this change?

src/Headers.php
```diff
@@ -74,7 +75,7 @@ public static function fromString($string)
 
             // check if a header name is present
             if (preg_match('/^(?P<name>[^()><@,;:\"\\/\[\]?={} \t]+):.*$/', $line, $matches)) {
-                if ($current) {
+                if (! empty($current)) {
```
Ocramius: Avoid using `empty()`, since it ignores non-existing variables too. Consider `=== ''` instead

src/PhpEnvironment/RemoteAddress.php
```diff
@@ -143,6 +143,7 @@ protected function getIpAddressFromProxy()
         // not know if it is a proxy server, or a client. As such, we treat it
         // as the originating IP.
         // @see http://en.wikipedia.org/wiki/X-Forwarded-For
+        /** @var string $ip */
```
Ocramius: Does it work if you map `/** @var string[] $ips */`?

src/PhpEnvironment/Response.php
```diff
@@ -96,7 +96,6 @@ public function sendHeaders()
             header($header->toString());
         }
 
-        $this->headersSent = true;
```
Ocramius: Is this on the parent class?

src/Request.php
```diff
@@ -274,6 +277,11 @@ public function getQuery($name = null, $default = null)
             return $this->queryParams;
         }
 
+        trigger_error(
```
Ocramius: Probably to be isolated into a separate patch

src/Request.php
```diff
@@ -307,18 +315,28 @@ public function getPost($name = null, $default = null)
             return $this->postParams;
         }
 
+        trigger_error(
```
Ocramius: Probably to be isolated into a separate patch

src/Request.php
```diff
@@ -307,18 +315,28 @@ public function getPost($name = null, $default = null)
...
+
         return $this->postParams->get($name, $default);
     }
 
     /**
      * Return the Cookie header, this is the same as calling $request->getHeaders()->get('Cookie');
      *
      * @convenience $request->getHeaders()->get('Cookie');
-     * @return Header\Cookie|bool
+     * @return Header\Cookie|false
```
Ocramius: `false` is not a type. Generally look for all `|false` or `false|` in your patch and replace them with `bool`

src/Request.php
```diff
@@ -351,6 +369,11 @@ public function getFiles($name = null, $default = null)
             return $this->fileParams;
         }
 
+        trigger_error(
```
Ocramius: To be isolated into a different patch

src/Request.php
```diff
@@ -366,13 +389,18 @@ public function getHeaders($name = null, $default = false)
...
             // this is only here for fromString lazy loading
-            $this->headers = (is_string($this->headers)) ? Headers::fromString($this->headers) : new Headers();
+            $this->headers = is_string($this->headers) ? Headers::fromString($this->headers) : new Headers();
         }
 
         if ($name === null) {
             return $this->headers;
         }
 
+        trigger_error(
```
Ocramius: To be isolated into a different patch

src/Request.php
```diff
@@ -386,11 +414,24 @@ public function getHeaders($name = null, $default = false)
      * @see Request::getHeaders()
      * @param string|null           $name            Header name to retrieve, or null to get the whole container.
      * @param mixed|null            $default         Default value to use when the requested header is missing.
-     * @return \Zend\Http\Headers|bool|\Zend\Http\Header\HeaderInterface|\ArrayIterator
+     * @return \Zend\Http\Headers|false|\Zend\Http\Header\HeaderInterface|\ArrayIterator
      */
     public function getHeader($name, $default = false)
     {
-        return $this->getHeaders($name, $default);
+        trigger_error(
```
Ocramius: To be isolated into a different patch

Ocramius (changes requested):

src/Cookies.php
thomasvargiu: [phpdoc](http://docs.phpdoc.org/guides/types.html#types-of-types) says:

> A type can either be
>
> - a class name, either Fully Qualified or aliases; such as \DateTime or Entity (with a namespace alias elsewhere in the same source file).
> - a keyword for a primitive in PHP, such as int or string.
> - a special keyword specific to the PHPDoc Standard, such as false or mixed.

So it should be a valid value

src/Cookies.php
thomasvargiu: I can't use that type definition, probably is better to check the ´$cpath` value and throw an exception if it isn't a `string`

src/Request.php
thomasvargiu: `false` keyword is accept by phpdoc

src/Cookies.php
thomasvargiu: I was looking around. Since PHP 7 `false` is internally a type.

There is a discussion in a phpstan [issue](https://github.com/phpstan/phpstan/issues/43#issuecomment-269314362).

Another resource of Nikic is this [article](https://nikic.github.io/2015/05/05/Internal-value-representation-in-PHP-7-part-1.html).

Ocramius: `false` is not a type: it is a value of the type `bool`, and hinting on a value is not something that tooling reading types should nor does deal with.

Whether PHP internally optimises `false` is a different (unrelated) story

thomasvargiu: Ok, I'm going to change it using `bool`, but I will need to add more type checks.

src/Cookies.php
```diff
@@ -316,7 +321,11 @@ protected function _matchPath($domains, $path)
         $ret = [];
 
         foreach ($domains as $dom => $pathsArray) {
-            foreach (array_keys($pathsArray) as $cpath) {
+            $keys = array_keys($pathsArray);
+            foreach ($keys as $cpath) {
+                if (! \is_string($cpath)) {
+                    throw new InvalidArgumentException('Domains is not a valid array');
+                }
```
weierophinney: I read the resolved conversation above, and have a few comments here:

- Why can't `$domains` be documented as `array<string, array<string, mixed>>`?
- Same note as previously about importing functions.
- This is new behavior and requires a corresponding test.

src/Exception/UnexpectedValueException.php
```diff
@@ -0,0 +1,26 @@
...
+     * @param mixed $actual
+     *
+     * @return UnexpectedValueException
+     */
+    public static function unexpectedType($expected, $actual)
+    {
+        return new static(sprintf(
+            'Expected %s. %s given',
+            $expected,
+            \is_object($actual) ? \get_class($actual) : \gettype($actual)
```
weierophinney: Either import these, or drop the global qualification.

src/Header/AbstractAccept.php
weierophinney: `stdClass` is not a type; it's a default implementation of an object in PHP. I agree with @Ocramius here.

src/Header/AbstractDate.php
```diff
@@ -97,7 +98,13 @@ public static function fromString($headerLine)
      */
     public static function fromTimeString($time)
     {
-        return static::fromTimestamp(strtotime($time));
+        $timestamp = strtotime((string) $time);
+
+        if (false === $timestamp) {
+            throw new InvalidArgumentException(sprintf('Invalid time %s', $time));
+        }
```
weierophinney: This is new behavior and requires a corresponding test.

src/Header/Accept/FieldValuePart/AbstractFieldValuePart.php
```diff
@@ -69,31 +69,40 @@ protected function getInternalValues()
...
+        return $this->raw;
+    }
+
+    /**
+     * @param string $key
+     * @return mixed
+     */
+    public function getInternalType($key)
+    {
+        return $this->getInternalValues()->$key;
```
weierophinney: This is a new method; why is it added?

src/Header/ContentType.php
weierophinney: No, because we're still targeting PHP 5.6 with the next minor release.

src/Header/Cookie.php
```diff
@@ -29,6 +29,13 @@ public static function fromSetCookieArray(array $setCookies)
                 ));
             }
 
+            if (null === $setCookie->getName()) {
+                throw new Exception\InvalidArgumentException(sprintf(
+                    '%s requires cookies with name',
+                    __METHOD__
+                ));
+            }
```
weierophinney: This is new behavior and requires a corresponding test.

src/Header/Cookie.php
```diff
@@ -55,10 +62,14 @@ public static function fromString($headerLine)
 
         $nvPairs = preg_split('#;\s*#', $value);
 
+        if (false === $nvPairs) {
+            throw new Exception\RuntimeException('Malformed Cookie header found');
+        }
```
weierophinney: This is new behavior and requires a corresponding test.

src/Header/SetCookie.php
```diff
@@ -486,18 +492,18 @@ public function setQuoteFieldValue($quotedValue)
...
         return $this->secure;
     }
 
     /**
-     * @param  bool $httponly
+     * @param  null|bool $httponly
      * @return SetCookie
      */
-    public function setHttponly($httponly)
+    public function setHttpOnly($httponly)
```
weierophinney: This is a breaking change; please revert, and also revert any calls to the method (I saw several above, but didn't realize it was because you'd renamed the method).

Alternatively, keep the original, and have it proxy to a new method named `setHttpOnly()` — though I'm not sure if PHP will allow it.

src/Header/SetCookie.php
```diff
@@ -507,9 +513,9 @@ public function setHttponly($httponly)
     }
 
     /**
-     * @return bool
+     * @return null|bool
      */
-    public function isHttponly()
+    public function isHttpOnly()
```
weierophinney: Same comment here as for `setHttponly()`; revert, or introduce the new method as a proxy to the original.

src/Headers.php
```diff
@@ -89,6 +90,10 @@ public static function fromString($string)
 
             if (preg_match("/^[ \t][^\r\n]*$/", $line, $matches)) {
                 // continuation: append to current line
+                if (! isset($current['line'])) {
+                    $current['line'] = '';
+                }
```
weierophinney: This likely needs a unit test.

src/PhpEnvironment/Request.php
```diff
@@ -93,6 +96,9 @@ public function getContent()
     {
         if (empty($this->content)) {
             $requestBody = file_get_contents('php://input');
+            if (false === $requestBody) {
+                throw new RuntimeException('Unable to get contents from php://input');
+            }
```
weierophinney: This is new behavior and requires a corresponding test.

src/Response.php
```diff
@@ -198,6 +202,9 @@ public static function fromString($string)
         if ($response->statusCode === static::STATUS_CODE_100) {
             $next = array_shift($lines); // take next line
             $next = empty($next) ? array_shift($lines) : $next; // take next or skip if empty
+            if (null === $next) {
+                throw new RuntimeException('Invalid response content');
+            }
```
weierophinney: This is new behavior and requires a corresponding test.

src/Response.php
```diff
@@ -608,8 +621,17 @@ protected function decodeDeflate($body)
...
-            return gzuncompress($body);
+            $uncompressed = gzuncompress($body);
+        } else {
+            $uncompressed = gzinflate($body);
+        }
+
+        if (false === $uncompressed) {
+            throw new Exception\RuntimeException(
+                'An error occurred during inflation'
+            );
```
weierophinney: This is new behavior and requires a corresponding test.

weierophinney (changes requested):
There's a lot of great changes here, but also a ton that look like potential backwards compatibility breaks and/or unrelated changes. Please review and address all comments from myself and @Ocramius, and we can then do a final review.

Thanks, @thomasvargiu !

weierophinney:
Removing the milestone; I suspect this patch will take a bit of back and forth before it's ready, and we're otherwise ready for the 2.9.0 release. We can target 2.10.0 for this.

weierophinney:
@thomasvargiu You will need to rebase against current develop before you start applying changes.

thomasvargiu:
> @thomasvargiu You will need to rebase against current develop before you start applying changes.

@weierophinney should I target develop branch for merge?

src/Cookies.php
thomasvargiu: There was already `false` return types in the code. I'll try to change them.

weierophinney:
Yes, please.

src/Client.php
```diff
@@ -219,7 +225,10 @@ public function getAdapter()
             $this->setAdapter($this->config['adapter']);
         }
 
-        return $this->adapter;
+        /** @var Client\Adapter\AdapterInterface $adapter */
+        $adapter = $this->adapter;
+
+        return $adapter;
```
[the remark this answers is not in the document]
thomasvargiu: I think we can't say that `$adapter` can only be an `AdapterInterface` because is not set in constructor. `$adapter` will be `null` until some method will call `getAdapter()`, then a developer should know that `$this->adapter` can be `null`.

Maybe, can we initialize it in constructor?

src/ClientStatic.php
```diff
@@ -56,7 +58,9 @@ public static function get($url, $query = [], $headers = [], $body = null, $clie
         }
 
         if (! empty($headers) && is_array($headers)) {
-            $request->getHeaders()->addHeaders($headers);
+            /** @var Headers $headers */
+            $headers = $request->getHeaders();
```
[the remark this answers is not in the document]
thomasvargiu: Good catch!

src/Client.php
```diff
@@ -829,6 +854,8 @@ protected function calcAuthDigest($user, $password, $type = self::AUTH_BASIC, $d
                         );
                     }
                     $ha2 = md5($this->getMethod() . ':' . $this->getUri()->getPath() . ':' . md5($entityBody));
+                } else {
+                    throw new InvalidArgumentException('Invalid DIGEST auth data');
```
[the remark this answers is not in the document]
thomasvargiu: There are no tests for that part. It's not used in the code but it's exposed because it's a protected method.

It's a little bit complicated to test it for me, I just added the check because `$ha2` will not be defined.

If we need tests, I think I need help (or time) on it.

src/Client.php
```diff
@@ -850,8 +877,11 @@ protected function calcAuthDigest($user, $password, $type = self::AUTH_BASIC, $d
      */
     public function dispatch(Stdlib\RequestInterface $request, Stdlib\ResponseInterface $response = null)
     {
-        $response = $this->send($request);
-        return $response;
+        if (! $request instanceof Request) {
+            throw UnexpectedValueException::unexpectedType(Request::class, $request);
+        }
+
+        return $this->send($request);
```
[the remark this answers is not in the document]
thomasvargiu: Method `send()` can only accept an instance of `Request`. So I prefer the 2nd option.

src/Client.php
```diff
@@ -917,23 +947,23 @@ public function send(Request $request = null)
...
-            $secure = $uri->getScheme() == 'https';
+            $secure = $uri->getScheme() === 'https';
+
+            if (null === $uri->getHost()) {
+                throw new InvalidArgumentException('Invalid URI in request');
+            }
 
             // cookies
-            $cookie = $this->prepareCookies($uri->getHost(), $uri->getPath(), $secure);
+            $cookie = $this->prepareCookies($uri->getHost(), $uri->getPath() ?: '/', $secure);
```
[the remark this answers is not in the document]
thomasvargiu: Yes of course, and this is also a bug fix. When someone set an `Uri\Http()` with no path, cookies was not set.

src/Client.php
```diff
@@ -951,19 +981,35 @@ public function send(Request $request = null)
...
+
                 if ($stream === null) {
                     $stream = $this->getStream();
                     if (! is_resource($stream) && is_string($stream)) {
-                        $stream = fopen($stream, 'r');
+                        $stream = fopen($stream, 'rb');
                     }
                 }
+
+                if (! \is_resource($stream)) {
```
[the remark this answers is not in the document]
thomasvargiu: I'll remove it, I think it would be better to do it in another branch.

src/Client.php
```diff
@@ -951,19 +981,35 @@ public function send(Request $request = null)
...
+                    throw new UnexpectedValueException('Stream is not a resource');
```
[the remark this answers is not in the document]
thomasvargiu: This is really difficult to test. Honestly I think that `$stream` can only be a resource at this line, and previous lines too are useless. But I'm not so confident to remove them.

src/Client.php
```diff
@@ -951,19 +981,35 @@ public function send(Request $request = null)
...
                     rewind($stream);
                 }
+
                 // cleanup the adapter
                 $adapter->setOutputStream(null);
                 $response = Response\Stream::fromStream($response, $stream);
+
+                // streamName can just be a string at this point
+                if (! \is_string($this->streamName)) {
+                    throw new UnexpectedValueException('Unexpected value for streamName');
```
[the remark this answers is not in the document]
thomasvargiu: It's not a possible behavior,  so it can't be tested. That's why is an `UnexpectedValueException`.

We can ignore it on phpstan or we can override the value

src/Client.php
```diff
@@ -485,7 +502,7 @@ public function setParameterGet(array $query)
      * Reset all the HTTP parameters (request, response, etc)
      *
      * @param  bool   $clearCookies  Also clear all valid cookies? (defaults to false)
-     * @param  bool   $clearAuth     Also clear http authentication? (defaults to true)
+     * param  bool   $clearAuth     Also clear http authentication? (defaults to true)
```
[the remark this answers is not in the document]
thomasvargiu: Oh, this is because the argument is commented too. I'll remove both.

src/Client.php
```diff
@@ -975,15 +1021,20 @@ public function send(Request $request = null)
 
             // Get the cookies from response (if any)
             $setCookies = $response->getCookie();
-            if (! empty($setCookies)) {
+            if (false !== $setCookies && ! empty($setCookies)) {
```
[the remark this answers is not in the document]
thomasvargiu: Since `getCookies()` will return `bool` types too, I need to exclude them from conditions, so I added `if (! is_bool($setCookies) .....`

src/Client.php
```diff
@@ -1255,14 +1311,17 @@ protected function prepareBody()
...
+
+        if (! $headers->has('Content-Type')) {
             $hasFiles = ! empty($this->getRequest()->getFiles()->toArray());
             // If we have files to upload, force encType to multipart/form-data
             if ($hasFiles) {
                 $this->setEncType(self::ENC_FORMDATA);
             }
         } else {
-            $this->setEncType($this->getHeader('Content-Type'));
+            $this->setEncType($this->getHeader('Content-Type') ?: null);
```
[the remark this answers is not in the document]
thomasvargiu: Yes, here is `null`, but default value is an empty string supporting (probably) my assertion that we can drop the nullable value. Check my previous answer.

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
     - COMPOSER_ARGS="--no-interaction"
     - COVERAGE_DEPS="php-coveralls/php-coveralls"
     - TESTS_ZEND_HTTP_CLIENT_ONLINE=true
+    - PHPSTAN_DEPS="phpstan/phpstan:^0.10.3"
```
webimpress: In other repositories we are adding it to composer dev requirements. Maybe we should do the same here?
Then it will be much easier to run it locally. See: https://github.com/zendframework/zend-expressive/blob/master/composer.json#L40

We can also update to version 0.11 :)

src/PhpEnvironment/Response.php
thomasvargiu: No, it's not used anywhere, it's just a dynamic assignment with no uses, there is a method for that

src/Client.php
```diff
@@ -313,25 +322,33 @@ public function setUri($uri)
...
             $this->getRequest()->setUri($uri);
 
             // if host changed, the HTTP authentication should be cleared for security
             // reasons, see #4215 for a discussion - currently authentication is also
             // cleared for peer subdomains due to technical limits
             $nextHost = $this->getRequest()->getUri()->getHost();
+
+            if (null === $nextHost) {
+                throw new InvalidArgumentException('Invalid URI provided. No host found');
+            }
```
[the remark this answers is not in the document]
thomasvargiu: Tests added

src/Client.php
```diff
@@ -407,7 +424,7 @@ public function getArgSeparator()
     {
         $argSeparator = $this->config['argseparator'];
         if (empty($argSeparator)) {
-            $argSeparator = ini_get('arg_separator.output');
+            $argSeparator = ini_get('arg_separator.output') ?: '&';
```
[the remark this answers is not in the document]
thomasvargiu: Tests added

src/Client.php
```diff
@@ -498,7 +515,7 @@ public function resetParameters($clearCookies = false /*, $clearAuth = true */)
         $uri = $this->getUri();
 
         $this->streamName      = null;
-        $this->encType         = null;
+        $this->encType         = '';
```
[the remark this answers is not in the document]
thomasvargiu: See previous comment

src/Client.php
```diff
@@ -720,7 +741,11 @@ protected function openTempStream()
             $this->streamName = tempnam(
                 isset($this->config['streamtmpdir']) ? $this->config['streamtmpdir'] : sys_get_temp_dir(),
                 Client::class
-            );
+            ) ?: null;
+
+            if (null === $this->streamName) {
+                throw new RuntimeException('Unable to create temporary name for stream');
+            }
```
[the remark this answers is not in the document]
thomasvargiu: Tests added

src/Client.php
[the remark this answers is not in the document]
thomasvargiu: Covered by tests

src/Client.php
```diff
@@ -917,23 +947,23 @@ public function send(Request $request = null)
             // headers
             $headers = $this->prepareHeaders($body, $uri);
 
-            $secure = $uri->getScheme() == 'https';
+            $secure = $uri->getScheme() === 'https';
+
+            if (null === $uri->getHost()) {
+                throw new InvalidArgumentException('Invalid URI in request');
+            }
```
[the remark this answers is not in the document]
thomasvargiu: Covered by tests

src/Client.php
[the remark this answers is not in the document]
thomasvargiu: I changed the logic, asserting that Request's URI always have a default path `/`

src/Client.php
```diff
@@ -1432,27 +1493,33 @@ protected function flattenParametersArray($parray, $prefix = null)
      */
     protected function doRequest(Http $uri, $method, $secure = false, $headers = [], $body = '')
     {
+        if (null === $uri->getHost()) {
+            throw new InvalidArgumentException('Invalid URI');
+        }
```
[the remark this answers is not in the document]
thomasvargiu: Tests added

src/Client/Adapter/Socket.php
```diff
@@ -366,8 +368,17 @@ public function write($method, $uri, $httpVer = '1.1', $headers = [], $body = ''
...
-        $host = (strtolower($uri->getScheme()) == 'https' ? $this->config['ssltransport'] : 'tcp') . '://' . $host;
-        if ($this->connectedTo[0] != $host || $this->connectedTo[1] != $uri->getPort()) {
+        $scheme = $uri->getScheme();
+        $port = $uri->getPort();
+
+        if (null === $host || null === $scheme || null === $port) {
+            throw new AdapterException\InvalidArgumentException(
+                'Invalid Uri object'
+            );
+        }
```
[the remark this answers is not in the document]
thomasvargiu: `Uri` class is not part of this library, I'm going to add tests on it

src/Client/Adapter/Socket.php
```diff
@@ -422,14 +433,21 @@ public function read()
         $response = '';
         $gotStatus = false;
 
-        while (($line = fgets($this->socket)) !== false) {
+        if (null === $this->socket) {
+            throw new AdapterException\RuntimeException('Socket is not initialized');
+        }
+
+        $line = fgets($this->socket);
+        while (false !== $line) {
```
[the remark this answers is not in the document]
thomasvargiu: `phpstan` can't evaluate the assignment and the condition, so I preferred to put it outside.
There errors are:

```
 ------ -------------------------------------------------------------------------------
  Line   Client/Adapter/Socket.php
 ------ -------------------------------------------------------------------------------
  449    Parameter #1 $haystack of function strpos expects string, string|false given.
  452    Parameter #1 $str of function rtrim expects string, string|false given.
 ------ -------------------------------------------------------------------------------
```

src/Cookies.php
```diff
@@ -120,14 +121,16 @@ public function addCookie($cookie, $refUri = null)
      */
     public function addCookiesFromResponse(Response $response, $refUri)
     {
-        $cookieHdrs = $response->getHeaders()->get('Set-Cookie');
+        /** @var Headers $headers */
+        $headers = $response->getHeaders();
+        $cookieHdrs = $headers->get('Set-Cookie');
```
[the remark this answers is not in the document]
thomasvargiu: Done

src/Cookies.php
```diff
@@ -120,14 +121,16 @@ public function addCookie($cookie, $refUri = null)
...
 
-        if (is_array($cookieHdrs) || $cookieHdrs instanceof ArrayIterator) {
+        if ($cookieHdrs instanceof ArrayIterator) {
```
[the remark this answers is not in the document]
thomasvargiu: `Headers::get()` doesn't return array, only `ArrayIterator` for MultiHeader. Can you double check it?

src/Cookies.php
```diff
@@ -120,14 +121,16 @@ public function addCookie($cookie, $refUri = null)
...
             foreach ($cookieHdrs as $cookie) {
-                $this->addCookie($cookie, $refUri);
+                $this->addCookie($cookie);
             }
         } elseif (is_string($cookieHdrs)) {
-            $this->addCookie($cookieHdrs, $refUri);
+            $this->addCookie($cookieHdrs);
```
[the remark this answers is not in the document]
thomasvargiu: `$refUri` in `addCookie()` method is not used. But I reverted it to avoid BC breaks with extended classes

src/Cookies.php
```diff
@@ -213,14 +216,15 @@ public function getCookie($uri, $cookieName, $retAs = self::COOKIE_OBJECT)
         }
 
         // Get correct cookie path
-        $path = $uri->getPath();
+        $path = $uri->getPath() ?: '';
```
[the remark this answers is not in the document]
thomasvargiu: It will be fixed in next lines, but I changed it to have the same behaviour

src/Cookies.php
thomasvargiu: > Why can't `$domains` be documented as `array<string, array<string, mixed>>`?

Done. You're right, when we'll have type hint `SetCookie::matchCookiePath()` parameter will just accept a string.

src/Header/Accept/FieldValuePart/AbstractFieldValuePart.php
thomasvargiu: Removed. I was trying to avoid using of magic methods.

src/Header/Cookie.php
thomasvargiu: Removed and set `$nvPairs` as `string[]`. The only waay it could return ` false` is when pattern is not valid, and this can be covered in tests.

src/Header/SetCookie.php
thomasvargiu: Reverted. Methods are case insensitive.

src/Headers.php
thomasvargiu: Fixed and tested

src/PhpEnvironment/Request.php
thomasvargiu: Fixed and replaced with simpler condition

src/Response.php
thomasvargiu: Tests added

src/Response.php
thomasvargiu: Tests added

.travis.yml
thomasvargiu: Someone else told me to remove it and to add it in CI.

weierophinney:
This repository has been closed and moved to laminas/laminas-http; a new issue has been opened at https://github.com/laminas/laminas-http/issues/4.

weierophinney:
This repository has been moved to laminas/laminas-http. If you feel that this patch is still relevant, please re-open against that repository, and reference this issue. To re-open, we suggest the following workflow:
 - Squash all commits in your branch (`git rebase -i origin/{branch}`)
 - Make a note of all changed files (`git diff --name-only origin/{branch}...HEAD
 - Run the laminas/laminas-migration tool on the code.
 - Clone laminas/laminas-http to another directory.
 - Copy the files from the second bullet point to the clone of laminas/laminas-http.
 - In your clone of laminas/laminas-http, commit the files, push to your fork, and open the new PR.
 We will be providing tooling via laminas/laminas-migration soon to help  automate the process.

PR #158, opened 2018-10-17
Deprecated multiple behavior methods

thomasvargiu:
In order to improve code quality and easily add type hints in a future release, I started to deprecate some methods with multiple behavior and different return types based on arguments.

- `Zend\Http\Request::getHeader()` is deprecated
- `Zend\Http\Request::getHeaders()` with arguments is deprecated
- `Zend\Http\Request::getQuery()` with arguments is deprecated
- `Zend\Http\Request::getPost()` with arguments is deprecated
- `Zend\Http\Request::getFiles()` with arguments is deprecated

The deprecation is for methods used as a "proxy".  
Clients can access to them getting the collection first.

```php
$request->getHeader('header-name'); // deprecated
$request->getHeaders('header-name'); // deprecated

$request->getHeaders()->get('header-name'); // currently suggested access
```

```php
$request->getQuery('name'); // deprecated

$request->getQuery()->get('name'); // currently suggested access
```

```php
$request->getPost('name'); // deprecated

$request->getPost()->get('name'); // currently suggested access
```

```php
$request->getFiles('name'); // deprecated

$request->getFiles()->get('name'); // currently suggested access
```

This will allow to introduce type hints for these methods in the future , something like:

```php
use Zend\Stdlib\ParametersInterface;
use Zend\Http\Headers;

class Request
{
    public function getHeaders(): Headers;
    public function getQuery(): ParametersInterface;
    public function getPost(): ParametersInterface;
    public function getFiles(): ParametersInterface;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,11 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Deprecated
 
-- Nothing.
+- `Zend\Http\Request::getHeader()` is deprecated
+- `Zend\Http\Request::getHeaders()` with arguments is deprecated
+- `Zend\Http\Request::getQuery()` with arguments is deprecated
+- `Zend\Http\Request::getPost()` with arguments is deprecated
+- `Zend\Http\Request::getFiles()` with arguments is deprecated
 
 ### Removed
 
--- a/src/Request.php
+++ b/src/Request.php
@@ -274,6 +274,11 @@ class Request extends AbstractMessage implements RequestInterface
             return $this->queryParams;
         }
 
+        trigger_error(
+            'Getting specific query parameter is deprecated',
+            E_USER_DEPRECATED
+        );
+
         return $this->queryParams->get($name, $default);
     }
 
@@ -306,6 +311,11 @@ class Request extends AbstractMessage implements RequestInterface
         if ($name === null) {
             return $this->postParams;
         }
+
+        trigger_error(
+            'Getting specific post parameter is deprecated',
+            E_USER_DEPRECATED
+        );
 
         return $this->postParams->get($name, $default);
     }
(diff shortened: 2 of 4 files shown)


froschdesign:
@thomasvargiu 
I'm sorry, but in my opinion this pull request is unacceptable.

I see the following problems:

* an explanation is missing why this functionality should be removed
* the error messages do not contain an alternative way to get the values
* the documentation is not updated

And where is the plan or the pull requests to update components like zend-mvc?

thomasvargiu:
@froschdesign I was working on #157. In order to improve the code in a major release is useless to call 
```php
$request->getHeader('name');
```

when you can access the entire collection and get it with:

```php
$request->getHeaders()->get('name');
```

IMHO, having a different behavior and a different return values based on arguments is not a good code.  
Introducing this deprecation will help us to use strict types, reducing tests and making a better code.

Of course we can improve deprecation messages.

It is a deprecation, but we can remove that behavior only on a major release.

froschdesign:
@thomasvargiu 
I know the usage, but that's not the topic. You open a PR without any explanation, but with changes which generates warning messages.
Please add always a full explanation in the title and / or in the descriptions. That will help you and everyone else. Thanks!

---
 
The question also remains:

> And where is the plan or the pull requests to update components like zend-mvc?

thomasvargiu:
`zend-http` is still on version 2.x. `zend-mvc` currently requires `zend-http:^2.x`. We can work to improve `zend-http` and release a `3.x` version (maybe with type hints?). As usual, other libraries should be compliant with the supported dependencies.

Changes are easy because the same behavior exists in `zend-http:^2.x` and will exist in a future major release.

I'm going to improve the PR description.

weierophinney:
> `zend-http` is still on version 2.x. `zend-mvc` currently requires `zend-http:^2.x`. We can work to improve `zend-http` and release a `3.x` version (maybe with type hints?).

zend-mvc v4 will be more likely to use PSR-7 ([see thezend-mvc  roadmap document](https://discourse.zendframework.com/t/rfc-zend-mvc-4-design-changes/447) — if we even decide to work towards a v4 at this point), which leaves the future of zend-http pretty much in maintenance mode. I do not see any reason at this point to do deprecations when we are unlikely to release a new major version ever.

As such, I'm going to close this.

PR #161, opened 2018-11-13
Do not upper case first character of headers in socket adapter

### LINKED ISSUE #160: Passed headers are upper cased on first character by default
- [x] I was not able to find an [open](https://github.com/zendframework/zend-http/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-http/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/).)

The passed headers are being made ucfirst() by default. Which is unnecessary according to HTTP 1.1 (https://www.w3.org/Protocols/rfc2616/rfc2616.html and which was not changed in https://tools.ietf.org/html/rfc7230#appendix-A.2).

See:
https://github.com/zendframework/zend-http/blob/master/src/Client/Adapter/Socket.php#L384

We are facing an issue were an API provider is checking for a case sensitive x- header (which of course is not HTTP 1.1 compliant).

### Code to reproduce the issue

```php
$config = [
    'adapter' => 'Zend\Http\Client\Adapter\Socket',
    'ssltransport' => 'tls'
];
$client = new Client('http://httpbin.org', $config);
$request = new Request();
$request->setUri('http://httpbin.org/get');
$request->setHeaders(Headers::fromString('x-my-test: myTest'));
$response = $client->send($request);
```

### Expected results

The header parameter should have been the way it has been given:
x-my-test: myTest

### Actual results
The header is being upper cased for the first char resulting in:
X-my-test: myTest

DaDeather:
This PR Fixes #160 which causes headers being upper cased for their first character in key which causes a fail request to a non HTTP 1.1 compliant API.

HTTP 1.1 defines that the headers should be case insensitive. This means there is also no reason for upper casing the first character.

- [x] Are you fixing a bug?
The first character in a headers "name"/"key" was being upper cased in default for Socket and Test adapter.
The headers using the Socket and Test adapter should not be be upper cased for the first character any more.

### DIFF
--- a/src/Client/Adapter/Socket.php
+++ b/src/Client/Adapter/Socket.php
@@ -381,7 +381,7 @@ class Socket implements HttpAdapter, StreamInterface
         $request = $method . ' ' . $path . ' HTTP/' . $httpVer . "\r\n";
         foreach ($headers as $k => $v) {
             if (is_string($k)) {
-                $v = ucfirst($k) . ': ' . $v;
+                $v = $k . ': ' . $v;
             }
             $request .= $v . "\r\n";
         }
--- a/src/Client/Adapter/Test.php
+++ b/src/Client/Adapter/Test.php
@@ -132,7 +132,7 @@ class Test implements AdapterInterface
         $request = $method . ' ' . $path . ' HTTP/' . $httpVer . "\r\n";
         foreach ($headers as $k => $v) {
             if (is_string($k)) {
-                $v = ucfirst($k) . ': ' . $v;
+                $v = $k . ': ' . $v;
             }
             $request .= $v . "\r\n";
         }
--- a/test/Client/SocketTest.php
+++ b/test/Client/SocketTest.php
@@ -310,6 +310,24 @@ class SocketTest extends CommonHttpTests
     }
 
     /**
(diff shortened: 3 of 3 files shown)


tptrixtop:
May be make this feature optional, like property in class by default use ucfirst and if you need you could change that behaviour?)

DaDeather:
> May be make this feature optional, like property in class by default use ucfirst and if you need you could change that behaviour?)

I can't tell if it would make sense to have a property to configure whatever it should be upper cased or not. The only thing I can say is that its already not being modified for the Curl and Proxy adapter.
So I would assume that either this (the Socket and Test adapter) or the behaviour of Curl and Proxy adapter is not the desired one.

tptrixtop:
That make sense i think, i was just thinking about back compatibility, but that make no sense in that case)

also some test are not successful after your changes, i think you need investigate test report)

weierophinney:
Thanks, @DaDeather; merged to develop for release with 2.9.0.

PR #165, opened 2018-12-25
Detecting base url for console requests

webimpress:
Alternative solution for: https://github.com/zendframework/zend-test/pull/66

Related issue: https://github.com/zendframework/zend-test/issues/71

In `zend-test` on dispatch from test global variables are set and the base url is detected wrongly in `zend-http`. Please see test case for more details.

- [x] Are you fixing a bug?


weierophinney:
Thanks, @webimpress!

PR #166, opened 2018-12-25
Added PHP 7.3 support - updated Travis CI configuration

webimpress:

PR #167, opened 2018-12-25
Fixed tests on PHP 7.2+

webimpress:
DateTime incorporates microseconds
Tests were failing on PHP 7.2+.
See #166 


weierophinney:
Ironically, when I run this locally on 7.2, I get the same behavior as previously. That said, I'm on an older 7.2 version.

Thanks for the fix, @webimpress !

Issue #172, opened 2019-02-14
parseStatusLine throws exception for HTTP/2 Response


av3:
I want to write a small function to test, if a specific Host supports HTTP/2 - using zend-http and the CURL adapter with the CURL_HTTP_VERSION_2_0 option. But this ends in an exception:

> Zend \ Http \ Exception \ InvalidArgumentException
> A valid response status line was not found in the provided string

It seems that the regex within parseStatusLine (https://github.com/zendframework/zend-http/blob/master/src/Response.php#L252) doesn't match with the HTTP/2 response.

### Code to reproduce the issue (incomplete)

```php
$client = new \Zend\Http\Client\Client();
$options = [
    'adapter'      => \Zend\Http\Client\Adapter\Curl::class,
    'curloptions'  => [
    // ...
        CURLOPT_HTTP_VERSION   => CURL_HTTP_VERSION_2_0,
    ],
];
$client->setOptions($options);

// ...

$response = $client->send();
```

Test case for this:

```php
    public function testResponseFactoryFromStringCreatesValidHTTP2Response()
    {
        $string = 'HTTP/2.0 200 OK' . "\r\n\r\n" . 'Foo Bar';
        $response = \Zend\Http\Response::fromString($string);
        $this->assertEquals(200, $response->getStatusCode());
        $this->assertEquals('Foo Bar', $response->getContent());
    }
```

PR #173, opened 2019-02-14
Feature/http2 support

av3:
As described in #172 there is currently no support for HTTP/2. I tried to fix it on my own and was able to process a HTTP/2 response. Hopefully this PR makes sense; I also added/extended unit tests and the documentation.


src/AbstractMessage.php
```diff
@@ -34,16 +35,16 @@ abstract class AbstractMessage extends Message
...
      *
-     * @param  string $version (Must be 1.0 or 1.1)
+     * @param  string $version (Must be 1.0, 1.1 or 2)
      * @return AbstractMessage
      * @throws Exception\InvalidArgumentException
      */
     public function setVersion($version)
     {
-        if ($version != self::VERSION_10 && $version != self::VERSION_11) {
+        if ($version != self::VERSION_10 && $version != self::VERSION_11 && $version != self::VERSION_2) {
```
samsonasik: I think it can use `! in_array`:

```suggestion
        if (! in_array($version, [self::VERSION_10, self::VERSION_11, self::VERSION_2], true)) {
```

Issue #176, opened 2019-04-07
Proposal: Feature Policy (a new "secure header")


MadCat34:
The HTTP **Feature-Policy** header provides a mechanism to allow and deny the use of browser features in its own frame, and in iframes that it embeds.

RFC is currently in Draft, but it could be interesting: https://w3c.github.io/webappsec-feature-policy/

If it is OK, I will provide a PR.

MadCat34:
Feature Policy and Content Security Policy will need to be refactored (duplicate code: getDirectives() and setDirectives()).

MadCat34:
I have created a new abstract class AbstractDirectiveBasedHeader to move duplicates.
Not sure about the naming...

What do you think ?

Ocramius:
I'd probably avoid abstracting through inheritance: as soon as an RFC changes in an uncommon way, that means trouble.

MadCat34:
Ok, I wasn't sure it was necessary.
This can be done later.

PR #177, opened 2019-04-10
Add Feature Policy header

MadCat34:
The HTTP Feature-Policy header provides a mechanism to allow and deny the use of browser features in its own frame, and in iframes that it embeds.

RFC is currently in Draft, but it could be interesting: https://w3c.github.io/webappsec-feature-policy/


webimpress:
Resolves #176

michalbundyra:
Thanks, @MadCat34!

PR #182, opened 2019-08-12
$argv[0] can be empty

eweso:
In some cases $argv[0] can be empty and strpos will throw an "Empty needle" error. Therefore on line 493 $argv should be checked, that it is not empty and a string, instead of just checking its existence with isset().

Someone tried to attack one of my websites with "/index.php?++++hot=1&++++kw=%E8%93%9D%E7%89%99%E8%80%B3%E6%9C%BA&r=l" and this request caused an exception which was logged and where I found the error.

Provide a narrative description of what you are trying to accomplish:

- [x] Are you fixing a bug?
  - [ ] Detail the original, incorrect behavior.
  - [ ] Detail the new, expected behavior.
  - [ ] Base your feature on the `master` branch, and submit against that branch.
  - [ ] Add a regression test that demonstrates the bug, and proves the fix.

- [ ] Are you creating a new feature?
  - [ ] Why is the new feature needed? What purpose does it serve?
  - [ ] How will users use the new feature?
  - [ ] Base your feature on the `develop` branch, and submit against that branch.
  - [ ] Add only one feature per pull request; split multiple features over multiple pull requests
  - [ ] Add tests for the new feature.
  - [ ] Add documentation for the new feature.
  - [ ] Add a `CHANGELOG.md` entry for the new feature.

- [ ] Is this related to quality assurance?
  

- [ ] Is this related to documentation?

### DIFF
--- a/src/PhpEnvironment/Request.php
+++ b/src/PhpEnvironment/Request.php
@@ -490,7 +490,7 @@ class Request extends HttpRequest
             // matching PHP_SELF.
 
             $argv = $this->getServer()->get('argv', []);
-            if (!empty($argv[0]) && is_string($argv[0]) && strpos($filename, $argv[0]) === 0) {
+            if (! empty($argv[0]) && is_string($argv[0]) && strpos($filename, $argv[0]) === 0) {
                 $filename = substr($filename, strlen($argv[0]));
             }
 
--- a/src/PhpEnvironment/Request.php
+++ b/src/PhpEnvironment/Request.php
@@ -489,9 +489,13 @@ class Request extends HttpRequest
             // Backtrack up the SCRIPT_FILENAME to find the portion
             // matching PHP_SELF.
 
-            $argv = $this->getServer()->get('argv', []);
-            if (! empty($argv[0]) && is_string($argv[0]) && strpos($filename, $argv[0]) === 0) {
-                $filename = substr($filename, strlen($argv[0]));
+            // Only for CLI requests argv[0] contains script filename
+            // @see https://www.php.net/manual/en/reserved.variables.server.php
+            if (PHP_SAPI === 'cli') {
+                $argv = $this->getServer()->get('argv', []);
+                if (strpos($filename, $argv[0]) === 0) {
+                    $filename = substr($filename, strlen($argv[0]));
+                }
             }
 
             $baseUrl  = '/';
--- a/src/PhpEnvironment/Request.php
+++ b/src/PhpEnvironment/Request.php
@@ -490,7 +490,7 @@ class Request extends HttpRequest
             // matching PHP_SELF.
 
             $argv = $this->getServer()->get('argv', []);
-            if (isset($argv[0]) && strpos($filename, $argv[0]) === 0) {
+            if (!empty($argv[0]) && is_string($argv[0]) && strpos($filename, $argv[0]) === 0) {
                 $filename = substr($filename, strlen($argv[0]));
             }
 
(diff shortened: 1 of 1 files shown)


src/PhpEnvironment/Request.php
```diff
@@ -490,7 +490,7 @@ protected function detectBaseUrl()
             // matching PHP_SELF.
 
             $argv = $this->getServer()->get('argv', []);
-            if (isset($argv[0]) && strpos($filename, $argv[0]) === 0) {
+            if (!empty($argv[0]) && is_string($argv[0]) && strpos($filename, $argv[0]) === 0) {
```
dennybrandes: Just simple fix, add space between ! and empty.

dennybrandes: Maybe you can also UnitTest for it?

eweso: Is there really a unit test needed? I just made the isset argument more strict. Strpos expects an not empty string as needle and I just ensure, that $argv[0] is always an non empty string.

dennybrandes: Just to cover the bug/case that it will not happen again after further changes.

dennybrandes: @webimpress what is your opinion?

webimpress:
Hi @eweso,

thanks for your contribution. I've checked it quickly and it looks like you have enabled
[`register_argc_argv`](http://www.php.net/manual/en/ini.core.php#ini.register-argc-argv) in your php.ini that you can see `argv`/`argc` in SERVER from `GET` (non-CLI) requests. This setting is disabled by default.

I've changed my configuration, tried the request you provided and the results of `SERVER['argv']` is as follows:
```
array (size=9)
  0 => string '' (length=0)
  1 => string '' (length=0)
  2 => string '' (length=0)
  3 => string '' (length=0)
  4 => string 'hot=1&' (length=6)
  5 => string '' (length=0)
  6 => string '' (length=0)
  7 => string '' (length=0)
  8 => string 'kw=%E8%93%9D%E7%89%99%E8%80%B3%E6%9C%BA&r=l' (length=43)
```

so yeah - `argv[0]` is empty.

Looking at the code, I don't think that we should really process `SERVER['argv']` there for GET request as it contains (per documentation: https://www.php.net/manual/en/reserved.variables.server.php) the query string.

We are detecting there the base url, so it cannot be detected from query string.

I would suggest to change your PHP configuration, unless you really need to use somewhere _SERVER['argv'] for GET requests.

src/PhpEnvironment/Request.php
webimpress: Hi, see my comment below. I think we need another fix - as argv shouldn't be used for non-CLI requests. And for CLI request, from what I have checked, it's not possible to have empty `argv[0]`.

michalbundyra:
As I said in previous comment - we don't really want process argv, as for GET requests it contains query string only - so we shouldn't use it for script filename. It should be used only in CLI mode.

Can't see a nice way to test it for GET request, for CLI test is in place already.

src/PhpEnvironment/Request.php
```diff
@@ -489,9 +489,13 @@ protected function detectBaseUrl()
             // Backtrack up the SCRIPT_FILENAME to find the portion
             // matching PHP_SELF.
 
-            $argv = $this->getServer()->get('argv', []);
-            if (isset($argv[0]) && strpos($filename, $argv[0]) === 0) {
-                $filename = substr($filename, strlen($argv[0]));
+            // Only for CLI requests argv[0] contains script filename
+            // @see https://www.php.net/manual/en/reserved.variables.server.php
+            if (PHP_SAPI === 'cli') {
```
weierophinney: :+1: This makes sense; only test argv if we're in a CLI request.

michalbundyra:
Thanks, @eweso!

PR #183, opened 2019-08-18
reduce getHeaders() call in Request::setServer()

samsonasik:

### DIFF
--- a/src/PhpEnvironment/Request.php
+++ b/src/PhpEnvironment/Request.php
@@ -233,7 +233,7 @@ class Request extends HttpRequest
             }
         }
 
-        $headers = $this->getHeaders()->addHeaders($headers);
+        $this->getHeaders()->addHeaders($headers);
 
         // set method
         if (isset($this->serverParams['REQUEST_METHOD'])) {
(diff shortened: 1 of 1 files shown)


src/PhpEnvironment/Request.php
```diff
@@ -233,7 +233,7 @@ public function setServer(ParametersInterface $server)
             }
         }
 
-        $this->getHeaders()->addHeaders($headers);
+        $headers = $this->getHeaders()->addHeaders($headers);
```
webimpress: Personally I am not sure if it is good improvement, I mean, for me it is less readable - as definition of the variable and usage of it is several lines below.
Also it takes advantages of fluent interfaces, and as far as I know we tend to drop them.

src/PhpEnvironment/Request.php
```diff
@@ -266,8 +266,8 @@ public function setServer(ParametersInterface $server)
         $port = null;
 
         // Set the host
-        if ($this->getHeaders()->get('host')) {
-            $host = $this->getHeaders()->get('host')->getFieldValue();
+        if ($headers->get('host')) {
```
webimpress: We have then twice `->get('host')`.
What about:
```php
$headerHost = $this->getHeaders()->get('host');
if ($headerHost) {
    $host = $headerHost->getFieldValue();
    ...
```

src/PhpEnvironment/Request.php
samsonasik: Ok, I reverted the change of re-assignment headers.

src/PhpEnvironment/Request.php
samsonasik: Ok, implemented assign first to `$headerHost`

webimpress:
Thanks, @samsonasik!

PR #184, opened 2019-08-21
process the response from a proxy with "HTTP/1.1 200 Connection established" header line

yugeon:
Fixing a bug when the response from a proxy contains first line as "HTTP/1.1 200 Connection established"


webimpress:
@yugeon Thanks for your contribution. Would you be able to provide unit test to cover the change, please?

michalbundyra:
Thanks, @yugeon!

Issue #174, opened 2019-02-19
Implement SameSite directive for SetCookie


jroedel:
I would like to use the [SameSite directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie) when setting cookies in ZF3, but there doesn't seem to be an option to do so.

This should be a 10th parameter for the SetCookie constructor I guess.

### Code to reproduce the issue (example constructor call)

```php
$setCookie = new SetCookie($cookieName, $locale, null, $path, null, $secure, $httpOnly, $maxAge, $version, SetCookie::SAME_SITE_STRICT);
```

### Expected results

This should set a cookie with the indicated SameSite directive

### Actual results

Doesn't allow for specification

carnage:
A possible temp band aid solution might look like:
```
        $cookie = new class(
//cookie options
        ) extends SetCookie {
            public function getFieldValue()
            {
                return parent::getFieldValue() . '; SameSite=lax';
            }
        };
```

PR #186, opened 2019-09-25
Add SameSite directive #174

MadCat34:
This PR allow to add the SameSite directive in SetCookie header

See #174

### DIFF
--- a/src/Header/SetCookie.php
+++ b/src/Header/SetCookie.php
@@ -31,7 +31,7 @@ class SetCookie implements MultipleHeaderInterface
     const SAME_SITE_LAX = 'Lax';
 
     /**
-     * Vookie will be sent with same-site and cross-site requests.
+     * Cookie will be sent with same-site and cross-site requests.
      */
     const SAME_SITE_NONE = 'None';
 
@@ -231,7 +231,7 @@ class SetCookie implements MultipleHeaderInterface
      * @param   bool                $httponly
      * @param   string              $maxAge
      * @param   int                 $version
-     * @param   string              $sameSite
+     * @param   string|null         $sameSite
      */
     public function __construct(
         $name = null,
@@ -334,7 +334,7 @@ class SetCookie implements MultipleHeaderInterface
         }
 
         $sameSite = $this->getSameSite();
-        if ($sameSite !== null && in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+        if (in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
             $fieldValue .= '; SameSite=' . $sameSite;
         }
 
@@ -621,11 +621,15 @@ class SetCookie implements MultipleHeaderInterface
     }
 
     /**
-     * @param string $sameSite
+     * @param string|null $sameSite
      * @return SetCookie
+     * @throws Exception\InvalidArgumentException
      */
     public function setSameSite($sameSite)
     {
+        if (!in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+            throw new Exception\InvalidArgumentException('"' . $sameSite . '" is not an allowed value (Strict, Lax, None) for SameSite directive.');
+        }
         $this->sameSite = $sameSite;
         return $this;
     }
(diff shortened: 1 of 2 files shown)


MadCat34:
Hi @webimpress 
Thanks for your return.

I have added a constant for None value.

src/Header/SetCookie.php
```diff
@@ -17,6 +17,29 @@
...
+
+    /**
+     * Cookie will not be passed for any cross-domain requests unless it's a regular link that navigates user
+     * to the target site.
+     * Other requests methods (such as POST and PUT) and XHR requests will not contain this cookie.
+     */
+    const SAME_SITE_LAX = 'Lax';
+
+    /**
+     * Vookie will be sent with same-site and cross-site requests.
```
webimpress: Typo: cookie

src/Header/SetCookie.php
```diff
@@ -17,6 +17,29 @@
...
+     */
+    const SAME_SITE_NONE = 'None';
+
+    const SAME_SITE_ALLOWED_VALUES = [
+        self::SAME_SITE_STRICT,
+        self::SAME_SITE_LAX
```
webimpress: Should contain also NONE

src/Header/SetCookie.php
```diff
@@ -572,6 +611,24 @@ public function isSessionCookie()
...
+    /**
+     * @return string|null
+     */
+    public function getSameSite()
+    {
+        return $this->sameSite;
+    }
+
+    /**
+     * @param string $sameSite
```
webimpress: string or null

src/Header/SetCookie.php
```diff
@@ -572,6 +611,24 @@ public function isSessionCookie()
...
+     * @return SetCookie
+     */
+    public function setSameSite($sameSite)
+    {
+        $this->sameSite = $sameSite;
```
webimpress: I believe we should check here if we set the correct value, and throw exception if not?

src/Header/SetCookie.php
```diff
@@ -199,6 +230,7 @@ public static function fromString($headerLine, $bypassHeaderFieldName = false)
      * @param   bool                $httponly
      * @param   string              $maxAge
      * @param   int                 $version
+     * @param   string              $sameSite
```
webimpress: string|null, I see other types should be updated also, but this is out of scope

src/Header/SetCookie.php
```diff
@@ -298,6 +333,11 @@ public function getFieldValue()
             $fieldValue .= '; HttpOnly';
         }
 
+        $sameSite = $this->getSameSite();
+        if ($sameSite !== null && in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
```
webimpress: first condition is redundant, `in_array` is sufficient

src/Header/SetCookie.php
MadCat34: Done !

src/Header/SetCookie.php
MadCat34: Sure :-)

src/Header/SetCookie.php
```diff
@@ -572,6 +612,30 @@ public function isSessionCookie()
...
+     * @param string|null $sameSite
+     * @return SetCookie
+     * @throws Exception\InvalidArgumentException
+     */
+    public function setSameSite($sameSite)
+    {
+        if (! in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+            throw new Exception\InvalidArgumentException(
+                '"' . $sameSite . '" is not an allowed value (Strict, Lax, None) for SameSite directive.'
+            );
```
webimpress: Need unit test for that case, and we are good :) thx

src/Header/SetCookie.php
MadCat34: $sameSite !== null must be tested because if $sameSite is null, we don't want `; SameSite=` in header and null must not be in allowed value (it breaks later)

src/Header/SetCookie.php
MadCat34: ok :)

src/Header/SetCookie.php
webimpress: yeah, and as we can't set invalid value we don't need in_array check ;-)

src/Header/SetCookie.php
MadCat34: Done :-)

michalbundyra:
Thanks, @MadCat34!

PR #187, opened 2019-09-27
Throw exception when header is invalid

### LINKED ISSUE #180: Invalid header results in infinite recursion
- [x] I was not able to find an [open](https://github.com/zendframework/zend-http/issues?q=is%3Aopen) or [closed](https://github.com/zendframework/zend-http/issues?q=is%3Aclosed) issue matching what I'm seeing.
 - [x] This is not a question. (Questions should be asked on [chat](https://zendframework.slack.com/) ([Signup here](https://zendframework-slack.herokuapp.com/)) or our [forums](https://discourse.zendframework.com/).)

Hey,
I'm using Zend Framework 3 in one of my projects and I'm encountering an issue during redirection. I wanted to pass a query parameter that includes \r and \n. However, the server stops responding because Zend enters an infinite recursion loop in [this](https://github.com/zendframework/zend-http/blob/master/src/Headers.php#L475) line.

To avoid this issue, I'm URL encoding the query parameter.

### Code to reproduce the issue
(in some Controller)

```php
 return $this->redirect()
                ->toUrl('/mail-campaign/edit?testMailSent=false&message=test \r \n test');
```

### Expected results

Zend should throw an error if the header is invalid.

### Actual results
Zend is entering infinite loop.

michalbundyra:
Fixes #180 - infinite recursion

- [x] Are you fixing a bug?

### DIFF
--- a/src/Headers.php
+++ b/src/Headers.php
@@ -472,6 +472,10 @@ class Headers implements Countable, Iterator
         try {
             $headers = $class::fromString($current['line']);
         } catch (Exception\InvalidArgumentException $exception) {
+            if ($isGeneric) {
+                throw $exception;
+            }
+
             return $this->lazyLoadHeader($index, true);
         }
         if (is_array($headers)) {
--- a/test/HeadersTest.php
+++ b/test/HeadersTest.php
@@ -315,4 +315,14 @@ class HeadersTest extends TestCase
         $this->expectException(RuntimeException::class);
         Headers::fromString("Fake: foo-bar\r\n\r\nevilContent");
     }
+
+    public function testThrowExceptionOnInvalidHeader()
+    {
+        $headers = new Headers();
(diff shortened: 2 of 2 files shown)


roypalacios:
Hi @webimpress,

It looks good to me. I'll only add few comments:
`
            // Generic Header should throw an exception if it fails
            if ($isGeneric) {
                throw $exception;
            }

            // Retry one more time with GenericHeader
            return $this->lazyLoadHeader($index, true);
`
Also, there is a minor optimization we can make at the beginning of the file:

> If Generic is true, there is no need to call the ClassLoader.

webimpress:
@roypalacios 

Thanks for suggestions. I should be working on zend-http 'soon'. As you can see there is a lot of PR I've created recently and more is incoming. I just need to find a time to describe and merge all of them to issue new release. Thanks for your patience.

PR #195, opened 2019-10-15
Improvement: tests

webimpress:
- Remove redundant bootstrap file (we can use composer autoload)
- Optimize getting static files (`__DIR__` notation).
- Rename protected method in tests to avoid `_` prefix,
  changed also to private.
- Optimize if statements and getting values from env vars
- Add PHPDoc with property type


webimpress:
All of it is in preparation to run "online" tests. I have done it on my branch already, but want to push changes part by part, instead everything in one PR. 

@froschdesign Thanks for your reviews. Appreciate a lot!

PR #199, opened 2019-11-29
Fix: decompressed response saved into file when streaming

### LINKED ISSUE #91: Client, Socket adapter: compressed responses not decompressed when streaming
The default Client pathways that use the Socket adapter automatically handle gzipped and deflated response bodies, and therefore include them in the Accept-Encoding header outgoing on responses as long as the zlib extension is present.

This automatic handling doesn't extend to the streaming support, however. In addition, the Client doesn't account for this fact and will still send an Accept-Encoding header indicating deflate and zlib are acceptable. The result is that the server will obey the Accept-Encoding header and send a compressed response, none of the Client, the Adapter, or the Response will decompress it, and it will be streamed as-is, in compressed form.

Because of the workflow involved in using the client, calling code can't know to add a decompression stream filter, as it would have to inspect headers to determine if the response was in fact compressed, but by the time those headers are available, `send()` has already been called and the data has already been streamed to its destination. The best available workaround is to manually set an Accept-Encoding header to tell servers that compression is not supported when using streams.

The Client should either not send the Accept-Encoding header when setStream has been called, or the Client or Adapter should automatically attach an appropriate stream filter to decompress the response according to the response headers.

This appears to be the same issue as the longstanding ZF1-era issue [ZF-10878](https://framework.zend.com/issues/browse/ZF-10878). I'm not sure how or if it's related to #89 which is Curl-specific and doesn't require streaming.

michalbundyra:
- [x] Are you fixing a bug?

When we set stream we'd like to get decompressed resource in the file.

Fixes #91

/cc @zerocrates

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -1184,7 +1184,7 @@ class Client implements Stdlib\DispatchableInterface
         // Set the Accept-encoding header if not set - depending on whether
         // zlib is available or not.
         if (! $this->getRequest()->getHeaders()->has('Accept-Encoding')) {
-            if (function_exists('gzinflate')) {
+            if (empty($this->config['outputstream']) && function_exists('gzinflate')) {
                 $headers['Accept-Encoding'] = 'gzip, deflate';
             } else {
                 $headers['Accept-Encoding'] = 'identity';
--- a/test/ClientTest.php
+++ b/test/ClientTest.php
@@ -12,6 +12,9 @@ use PHPUnit\Framework\TestCase;
 use ReflectionMethod;
 use Zend\Http\Client;
 use Zend\Http\Client\Adapter\AdapterInterface;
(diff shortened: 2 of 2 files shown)


src/Client.php
```diff
@@ -1184,7 +1184,7 @@ protected function prepareHeaders($body, $uri)
         // Set the Accept-encoding header if not set - depending on whether
         // zlib is available or not.
         if (! $this->getRequest()->getHeaders()->has('Accept-Encoding')) {
-            if (function_exists('gzinflate')) {
+            if (empty($this->config['outputstream']) && function_exists('gzinflate')) {
```
weierophinney: I'm curious: can you explain why this is needed?

michalbundyra: See the test - we'd like to have uncompressed content stored in the file if we are using streams and we set the file.

Issue #206, opened 2019-12-29
SameSite directive implementation in Zend\Http\Header\SetCookie throws exception on lowercase values


wilcol:
Hi all,

When upgrading from zend-http 2.10 to 2.11 my REST integration with Exact stopt working. After an investigation it turned out that the newly implemented SameSite directive was the culprit. Its cookie did contain the SameSite element, but its value was in lowercase.

`string(117) "Set-Cookie: ExactServer{7d1b186a-964e-4536-b49e-f36d86d47892}=Division=XXXXXX; SameSite=lax; path=/; secure; HttpOnly"`

This is the exception message:
`"Invalid value provided for SameSite directive: "lax"; expected one of: Strict, Lax or None"`
It is however caught in `Zend\Http\Header::lazyLoadHeader()` and therefore never shown.

In `Zend\Http\Header\SetCookie::setSameSite()`, the in_array method is case sensitive for strings and the defined values all start with capitals (Strict, Lax, None)

I have found several draft specifications all dealing with case insensitive match of the value:
https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1 and https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-5.3.7

In the current situation I'm forced to remain on zend-http 2.10. My question is if a pull request allowing for case insensitive comparison would get accepted.

michalbundyra:
@wilcol Thanks for reporting the issue. It looks like the value of SameSite should be case-insensitive. Please submit the PR with the test case. Thanks!

wilcol:
Done in https://github.com/zendframework/zend-http/pull/207. Thank you.

PR #207, opened 2019-12-30
fixes case sensitivity for SameSite directive

wilcol:
This fix handles case sensitivity for SameSite directive in cookie headers.

Prior to this fix the SameSite values were case sensitive compared with their capitalized names (`Lax`, `Strict`, `None`). This fix allows for case insensitive comparison, allowing values like `lax` and `LAX`. 

See https://github.com/zendframework/zend-http/issues/206 for more information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- Nothing.
+- fixes case sensitivity for SameSite directive.
 
 ## 2.11.1 - 2019-12-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ All notable changes to this project will be documented in this file, in reverse 
 
 ### Fixed
 
-- fixes case sensitivity for SameSite directive.
+- [#207](https://github.com/zendframework/zend-http/pull/207) fixes case sensitivity for SameSite directive.
 
 ## 2.11.1 - 2019-12-04
 
--- a/src/Header/SetCookie.php
+++ b/src/Header/SetCookie.php
@@ -618,18 +618,17 @@ class SetCookie implements MultipleHeaderInterface
      */
     public function setSameSite($sameSite)
     {
-        if ($sameSite !== null) {
-            if (array_key_exists(strtolower($sameSite), self::SAME_SITE_ALLOWED_VALUES)) {
-                $this->sameSite = self::SAME_SITE_ALLOWED_VALUES[strtolower($sameSite)];
-            } else {
-                throw new Exception\InvalidArgumentException(sprintf(
-                    'Invalid value provided for SameSite directive: "%s"; expected one of: Strict, Lax or None',
-                    is_scalar($sameSite) ? $sameSite : gettype($sameSite)
-                ));
-            }
-        } else {
-            $this->sameSite = $sameSite;
+        if ($sameSite === null) {
+            $this->sameSite = null;
+            return $this;
         }
+        if (! array_key_exists(strtolower($sameSite), self::SAME_SITE_ALLOWED_VALUES)) {
+            throw new Exception\InvalidArgumentException(sprintf(
+                'Invalid value provided for SameSite directive: "%s"; expected one of: Strict, Lax or None',
+                is_scalar($sameSite) ? $sameSite : gettype($sameSite)
+            ));
+        }
+        $this->sameSite = self::SAME_SITE_ALLOWED_VALUES[strtolower($sameSite)];
         return $this;
     }
 
--- a/src/Header/SetCookie.php
+++ b/src/Header/SetCookie.php
@@ -39,9 +39,9 @@ class SetCookie implements MultipleHeaderInterface
      * @internal
      */
     const SAME_SITE_ALLOWED_VALUES = [
-        self::SAME_SITE_STRICT,
-        self::SAME_SITE_LAX,
-        self::SAME_SITE_NONE,
+        'strict',
+        'lax',
+        'none',
     ];
 
     /**
@@ -337,8 +337,7 @@ class SetCookie implements MultipleHeaderInterface
         }
 
         $sameSite = $this->getSameSite();
-        if ($sameSite !== null
-            && in_array(strtolower($sameSite), array_map('strtolower', self::SAME_SITE_ALLOWED_VALUES), true)) {
+        if ($sameSite !== null && in_array(strtolower($sameSite), self::SAME_SITE_ALLOWED_VALUES, true)) {
             $fieldValue .= '; SameSite=' . $sameSite;
         }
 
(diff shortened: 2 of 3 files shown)


src/Header/SetCookie.php
```diff
@@ -337,7 +337,8 @@ public function getFieldValue()
         }
 
         $sameSite = $this->getSameSite();
-        if ($sameSite !== null && in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+        if ($sameSite !== null
+            && in_array(strtolower($sameSite), array_map('strtolower', self::SAME_SITE_ALLOWED_VALUES), true)) {
```
michalbundyra: I would suggest to change the values in the constant `SAME_SITE_ALLOWED_VALUES` as it is marked as `internal`. Then we don't need to lowercase it here.

src/Header/SetCookie.php
```diff
@@ -251,15 +251,15 @@ public function __construct(
...
-             ->setSameSite($sameSite);
+            ->setValue($value)
+            ->setVersion($version)
+            ->setMaxAge($maxAge)
+            ->setDomain($domain)
+            ->setExpires($expires)
+            ->setPath($path)
+            ->setSecure($secure)
+            ->setHttpOnly($httponly)
+            ->setSameSite($sameSite);
```
michalbundyra: Please revert, the change is unrelated to the hotfix.

src/Header/SetCookie.php
```diff
@@ -618,7 +619,8 @@ public function getSameSite()
      */
     public function setSameSite($sameSite)
     {
-        if ($sameSite !== null && ! in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+        if ($sameSite !== null
+            && ! in_array(strtolower($sameSite), array_map('strtolower', self::SAME_SITE_ALLOWED_VALUES), true)) {
```
michalbundyra: As above - if we change values in the constant then we don't need to have array_map here.

test/Header/SetCookieTest.php
```diff
@@ -161,6 +187,30 @@ public function testSetCookieFromStringCanCreateSingleHeader()
         $this->assertTrue($setCookieHeader->isSecure());
         $this->assertTrue($setCookieHeader->isHttponly());
         $this->assertEquals(setCookie::SAME_SITE_STRICT, $setCookieHeader->getSameSite());
+        $this->assertEquals(
+            'myname=myvalue; Expires=Wed, 13-Jan-2021 22:23:01 GMT; Domain=docs.foo.com; '
+            . 'Path=/accounts; Secure; HttpOnly; SameSite=Strict',
+            $setCookieHeader->getFieldValue()
+        );
+
+        $setCookieHeader = SetCookie::fromString(
```
michalbundyra: It should be separate test case.
I see it is like that in that test case, but it's wrong...

CHANGELOG.md
```diff
@@ -44,7 +44,7 @@ All notable changes to this project will be documented in this file, in reverse
 
 ### Fixed
 
-- Nothing.
+- fixes case sensitivity for SameSite directive.
```
michalbundyra: Can be fixed on merge: it should contain link to the PR as we have for other changelogs

michalbundyra (changes requested):
@wilcol Thanks for your contribution. I have couple minor comments. In general it looks good. See my comments below. Thanks!

wilcol:
@michalbundyra, I processed your feedback. Thank you.

src/Header/SetCookie.php
```diff
@@ -39,9 +39,9 @@ class SetCookie implements MultipleHeaderInterface
      * @internal
      */
     const SAME_SITE_ALLOWED_VALUES = [
-        self::SAME_SITE_STRICT,
-        self::SAME_SITE_LAX,
-        self::SAME_SITE_NONE,
+        'strict',
+        'lax',
+        'none',
```
Xerkus: I would suggest it to be changed to
```php
 const SAME_SITE_ALLOWED_VALUES = [
        'strict' = self::SAME_SITE_STRICT,
        'lax' = self::SAME_SITE_LAX,
        'none' = self::SAME_SITE_NONE,
    ];
```
Lowercase assertion can be made against key and additionally that will allow to set canonical form value:
```
$this->sameSite = self::SAME_SITE_ALLOWED_VALUES[strtolower($sameSite)];
```

Xerkus: Considering that users might check value against constant, please make that change and add additional test ensuring case insensitive value is set to canonical form.

wilcol:
@Xerkus, I processed your feedback.

src/Header/SetCookie.php
```diff
@@ -618,13 +618,18 @@ public function getSameSite()
      */
     public function setSameSite($sameSite)
     {
-        if ($sameSite !== null && ! in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
-            throw new Exception\InvalidArgumentException(sprintf(
-                'Invalid value provided for SameSite directive: "%s"; expected one of: Strict, Lax or None',
-                is_scalar($sameSite) ? $sameSite : gettype($sameSite)
-            ));
+        if ($sameSite !== null) {
```
Xerkus: Early return trick can be used to increase readability by decreasing indentation level and avoiding else statements:
```php
if ($sameSite === null) {
    $this->sameSite = null;
    return;
}
```

Xerkus (changes requested):

Xerkus:
Please also add a separate test for asserting that SameSite value is canonicalized in setSameSite

wilcol:
@Xerkus all done!

src/Header/SetCookie.php
```diff
@@ -337,7 +337,7 @@ public function getFieldValue()
         }
 
         $sameSite = $this->getSameSite();
-        if ($sameSite !== null && in_array($sameSite, self::SAME_SITE_ALLOWED_VALUES, true)) {
+        if ($sameSite !== null && array_key_exists(strtolower($sameSite), self::SAME_SITE_ALLOWED_VALUES)) {
```
Xerkus: coding style: used functions must be imported with `use function` statements (under `use` block, separated by blank line)
Import just those you touched, that will be enough.

test/Header/SetCookieTest.php
```diff
@@ -105,6 +131,51 @@ public function testSetInvalidSameSiteDirectiveValueViaSetter()
...
+            'Wed, 13-Jan-2021 22:23:01 GMT',
+            '/accounts',
+            'docs.foo.com',
+            true,
+            true,
+            99,
+            9,
+            SetCookie::SAME_SITE_STRICT
+        );
+        $this->assertEquals(SetCookie::SAME_SITE_STRICT, $setCookieHeader->getSameSite());
```
Xerkus: Please also test `setSameSite()` explicitly. Just once would be enough.

Xerkus (changes requested):
Looks  good, just a bit of extra changes.

Xerkus (approved):
Perfect, thanks!

weierophinney:
As this was a bugfix, I rebased against master and merged there for a new 2.11.2 release.

Thanks, @wilcol !

